Issue #23, opened 2016-07-12
Open tools from results, not only their class


annkatrinkuessner:
As a user when I search for the name of a tool I want to be able to open that tool and not only its claas definition in order to use Algernon as a general purpose launcher.

# Acceptance Criteria
- [x] the `open` hotword can directly open a tool (rather than its class) additionally to it's existing functionality of opening Apps
  - A tool is any program listed in the Tools dropdown

# Old description
Algernon should also allow developers to open tools directly from the launcher (Browser, Workspace, Transcript, ...)

LeonMatthes:
Estimated effort: 5

LeonMatthes:
Currently not working in my image, needs further testing

BenBals:
Does not work in my image either.

BenBals:
I have written a test and (hopefully) a fix in #47.

PR #62, opened 2020-05-25
Implement #23: Open tools from results, not only their class

sinnaj-r:
The known Issues with the last implementation of #23 should be gone. Also some tests were added to prevent such issues from appearing again.


BenBals (approved):

BenBals (approved):

BenBals:
This pr only waits for #58, correct?

sinnaj-r:
> This pr only waits for #58, correct?

Yes.

Issue #60, opened 2020-05-24
Refactor ALGAction invocation


BenBals:
There is some black magic going on here.

```smalltalk
MenuItemMorph>>invokeWithEvent: evt
	"Perform the action associated with the given menu item."

	| w |
	self isEnabled ifFalse: [^ self].
	target class == HandMorph ifTrue: [(self notObsolete) ifFalse: [^ self]].
	owner ifNotNil:[self isStayUpItem ifFalse:[
		self flag: #workAround. "The tile system invokes menus straightforwardly so the menu might not be in the world."
		(w := self world) ifNotNil:[
			owner deleteIfPopUp: evt.
			"Repair damage before invoking the action for better feedback"
			w displayWorldSafely]]].
	selector ifNil:[^self].
	Cursor normal showWhile: [ | selArgCount |  "show cursor in case item opens a new MVC window"
		(selArgCount := selector numArgs) = 0
			ifTrue:
				[target perform: selector]
			ifFalse:
				[selArgCount = arguments size
					ifTrue: [target perform: selector withArguments: arguments]
					ifFalse: [target perform: selector withArguments: (arguments copyWith: evt)]]].
```

PR #64, opened 2020-05-26
Refactor #60: Refactor ALGAction invocation

sinnaj-r:
This PR removes the usage of the `doActionOnSelectedItem: Method.`
There are also some small test refactorings.


packages/Algernon.package/ALGAction.class/instance/do.st
```diff
@@ -0,0 +1,3 @@
+accessing
+do
+	self theTarget perform: self theSelector withArguments: self theArgs asArray
```
BenBals, line 3: Is there a way to remove metaprogramming here?

packages/AlgernonTests.package/ALGTest.class/instance/currentBrowsers.st
```diff
@@ -4,4 +4,4 @@ currentBrowsers
 	^ (ActiveWorld submorphs
 		select: [:each | (each isKindOf: SystemWindow)
 				and: [each model inheritsFromAnyIn: 
-					{Browser. PreferenceBrowser. MCWorkingCopyBrowser}]]) asSet
\ No newline at end of file
+					{Browser. PreferenceBrowser. MCWorkingCopyBrowser . TranscriptStream . MessageNames}]]) asSet
```
BenBals: Could you refactor this code to just close all newly opened `PluggableSystemWindow`s? That should do it without us having to manually specify each tool.

packages/AlgernonTests.package/ALGTest.class/instance/resultsClassIncludeName..st
```diff
@@ -0,0 +1,6 @@
+helpers
+resultsClassIncludeName: aNameString 
+	^ self algerInstance itemList items
```
BenBals, line 3: Use the helper `self items`

packages/AlgernonTests.package/ALGTest.class/instance/resultsIncludeName..st
```diff
@@ -0,0 +1,4 @@
+helpers
+resultsIncludeName: aNameString 
+	^ self algerInstance itemList items
```
BenBals, line 3: Use the helper `self items`.

BenBals (changes requested):
I have a few small code style things. Feel free to merge when you are done with reviewing my change requests.

packages/AlgernonTests.package/ALGTest.class/instance/resultsClassIncludeName..st
sinnaj-r: I didn't use it, because I tried to improve the performance of the FuzzyMatching tests. Therefore I did not want to put an additional collect (in form of the helper) in front of it.

BenBals: Fair enough.

packages/AlgernonTests.package/ALGTest.class/instance/currentBrowsers.st
sinnaj-r: Sure, I can do that.

packages/Algernon.package/ALGAction.class/instance/do.st
sinnaj-r: Hm, it's not that simple. The Types create an multiple Actions to perform code, which is found type classes. Therefore a bit of Metaprogramming is needed at least for the Actions and Names. We could move the "Action-Generation2 Part from the Types to Action Class Methods, but I'm not sure if that would really improve the current situation. 

If you have a different/better idea, feel free to share it with me. 🙂

BenBals: Could we just pass a block?

BenBals: So that each action has a block associated with it. I think that would make a lot of code nicer. Especially in places where actions need multiple arguments as this method only supports one argument.

sinnaj-r: Several arguments are also possible now. (As array). 
With blocks I worry that the "Debug-ability" would suffer, because you cannot look into blocks so easily.

BenBals: Is that a problem with blocks? I have not experienced that. I mean we're just building our own block, aren't we?

packages/AlgernonTests.package/ALGTest.class/instance/currentBrowsers.st
```diff
@@ -1,7 +1,4 @@
 helpers
 currentBrowsers
-
 	^ (ActiveWorld submorphs
-		select: [:each | (each isKindOf: SystemWindow)
-				and: [each model inheritsFromAnyIn: 
-					{Browser. PreferenceBrowser. MCWorkingCopyBrowser . TranscriptStream . MessageNames}]]) asSet
\ No newline at end of file
+		select: [:each | each isKindOf: SystemWindow]) asOrderedCollection
```
BenBals: What about ``SystemWindow allInstances``? Would that be faster?

packages/AlgernonTests.package/ALGTest.class/instance/currentBrowsers.st
sinnaj-r: I just committed the new "currentSystemWindows"-Method.

packages/Algernon.package/ALGAction.class/instance/do.st
sinnaj-r: Yes we are, but therefore we can't easily view the Arguments or the executed selector in the debugger (Since both are "embedded" in the block).
Although I have just learned that blocks at least have the method "sourceString", so it is at least possible with one "intermediate" step 🤔.

BenBals: But it's not super easy now either. And a block is probably pretty easy to trace to its definition.

BenBals: But I accept your judgement as this is your pr.

BenBals (approved):

BenBals:
Let me guess: We have to wait for #58?

PR #68, opened 2020-05-28
Fix and Refactor #2: long item names overlap with action

sinnaj-r:
This PR refactors the whole ALGResultItemMorph and ALGResultList from using MenuMorphs to normal Morphs. In this progress the bug #2 should be fixed too.


sinnaj-r:
At this moment the code is functional and #2 is fixed, but the code is not that clean; I'm gonna try to do some cleaning up over the next few days.

sinnaj-r:
But since I wrote a few more tests we have reached 95% test coverage 🎉

packages/Algernon.package/ALGIconFactory.class/class/iconSizeItemType.st
```diff
@@ -0,0 +1,4 @@
+constants
+iconSizeItemType
+
+	^ self iconSize * 0.75
```
BenBals: Why does this need to happen? Either the magic number needs its own constant (which I don't think) or you should add a comment.

packages/Algernon.package/ALGResultItemMorph.class/instance/createActionMorph.st
```diff
@@ -0,0 +1,9 @@
+initialization
+createActionMorph
+	^ TextMorph new
+		contentsWrapped: (self item defaultAlgeraction name asText addAttribute: TextAlignment rightFlush);
+		 textColor: self class actionTextColor;
+		
+		font: (TextStyle default fontOfSize: self class actionFontSize);
+		 hide;
+		 yourself
```
BenBals: Maybe this is just GitHub displaying the code wrongly (feel free to resolve this if that is the case) but there seems to be something weird going on with the indentation.

packages/Algernon.package/ALGResultItemMorph.class/instance/createHasChildrenMorph.st
```diff
@@ -0,0 +1,5 @@
+initialization
+createHasChildrenMorph
+	^ ((ALGIconFactory getInstance loadIconFor: #HasChildren)
```
BenBals: Could the singleton behaivior be hidden by directly sending `loadIconFor` to the class side of `ALGIconFactory`?

packages/Algernon.package/ALGResultItemMorph.class/instance/deselect.st
```diff
@@ -0,0 +1,6 @@
+event handling
+deselect
+	self color: self class normalFillStyle.
+
+	self hasChildrenMorph hide.
```
BenBals: `hasChildresMorph` sounds like this should be a boolean. If it is not: Please rename it.

packages/Algernon.package/ALGResultItemMorph.class/instance/hasChildrenMorph.st
```diff
@@ -0,0 +1,4 @@
+accessing
+hasChildrenMorph
```
BenBals: could you add a comment here why we need our own custom implementation instead of using submorphs?

packages/Algernon.package/ALGResultItemMorph.class/instance/select.st
```diff
@@ -0,0 +1,9 @@
+event handling
+select
+
+	self isTitle ifTrue: [ ^ self ].
```
BenBals, line 4: We should keep an eye on this. If there are more branching points, Title could be a subclass overwriting certain methods to do nothing.

packages/Algernon.package/ALGResultList.class/instance/updateWith..st
```diff
@@ -1,7 +1,8 @@
...
-updateWith: aCollection
-
-	self submorphs do: [ :submorph | self submorphs remove: submorph ].
-	aCollection copy do: [ :aCategory |
-		self addCategory: aCategory.
-		aCategory items do: [ :anAlgertem | self add: anAlgertem ] ]
\ No newline at end of file
+updateWith: aCollection 
+	self removeAllSubmorphs.
+	aCollection copy
```
BenBals: I dont see why you have to copy the collection here. Can you explain?

BenBals (approved):
There are two issues:

A screenshot

1. The font size for the action is to small on my hdpi screen.
2. The backdrop is completely white instead of a translucent shadow.

BenBals (changes requested):
I did not mean to approve. Hit the wrong button.

sinnaj-r:
> There are two issues:
> 
> A screenshot
> 
> 1. The font size for the action is to small on my hdpi screen.
> 2. The backdrop is completely white instead of a translucent shadow.

1. I could not find any possibilities for relative font sizes quickly. Are other parts of Squeak affected by this ?
2. Did you run the tests ? At least on my machine it's a side effect of problematic cleanup of tests; Therefore a `ALGInstallation instance cleanup` should resolve the issue. (AFAIK, this issue was not introduced by this branch)

packages/Algernon.package/ALGIconFactory.class/class/iconSizeItemType.st
sinnaj-r: Converted to a constant

packages/Algernon.package/ALGResultItemMorph.class/instance/createActionMorph.st
sinnaj-r: The Pretty Print is doing strange stuff, I manually aligned the code.

packages/Algernon.package/ALGResultItemMorph.class/instance/deselect.st
sinnaj-r: renamed to childrenIndicatorMorph

packages/Algernon.package/ALGResultItemMorph.class/instance/hasChildrenMorph.st
sinnaj-r: That was just very poor naming on my part. "hasChildrenMorph" is not a boolean, but the arrow morph that is rendered when an item has "sub" actions. (Now named childrenIndicatorMorph)

packages/Algernon.package/ALGResultList.class/instance/updateWith..st
sinnaj-r: It's part of legacy code. I just removed the copy and noticed no side effects so far.

packages/Algernon.package/ALGResultItemMorph.class/class/actionFontSize.st
```diff
@@ -1,3 +1,3 @@
 constants
 actionFontSize
-	^ 13
\ No newline at end of file
+	^ 8
```
BenBals, line 3: I complained it was to small, so you made it smaller?

BenBals (changes requested):
Unless there is some special vodoo going on with font sizes in Squeak, I suggest to increase the action text font size rather than decrease it.

packages/Algernon.package/ALGResultItemMorph.class/instance/createHasChildrenMorph.st
sinnaj-r: Done.

BenBals: (This is not really part of your issue. We can merge without it. If you decide not to implement it, open up an issue and resolve this coversation.)

BenBals:
> I could not find any possibilities for relative font sizes quickly. Are other parts of Squeak affected by this ?

Yes, everything in Squeak is quite small but still readable. This is just too small now.

BenBals:
> Did you run the tests ? At least on my machine it's a side effect of problematic cleanup of tests; Therefore a ALGInstallation instance cleanup should resolve the issue. (AFAIK, this issue was not introduced by this branch)

This seems to be the issue. Thank you. (We have #71 to solve this.)

BenBals (approved):
I now approve of the changes. A better solution of the font size thing would be great but we can merge without it. Same goes for the `ALGIconFactor` improvements.

packages/Algernon.package/ALGResultItemMorph.class/class/actionFontSize.st
sinnaj-r: I also changed the Font-Size adaption function (`TextStyle default fontOfPointSize`) to use points instead of "height".
But there seems to be some "voodoo" involved, because there seem to be fixed font sizes, so I can't scale them freely.

sinnaj-r:
> I now approve of the changes. A better solution of the font size thing would be great but we can merge without it. Same goes for the `ALGIconFactor` improvements.

Okay, thank you! If you think a complete refactoring of the icon class is necessary, feel free to create an issue for it. 
For the font size problem I have already created an issue.

Issue #17, opened 2016-07-07
display a cursor in search field


annkatrinkuessner:
As a user I want a curser in Algernon to make editing my request easier.

# Acceptance Criteria
- [x] have a curser in the algernon search field

PR #72, opened 2020-05-29
Cursor, implement #17

BenBals:


packages/Algernon.package/ALGInputField.class/instance/keyboardFocusColor.st
```diff
@@ -0,0 +1,3 @@
+accessing
+keyboardFocusColor
+	^ ALGBreadcrumbs defaultColor
```
sinnaj-r, line 4: Is that the correct color to use ?

packages/AlgernonTests.package/ALGTestGeneral.class/instance/testCloseWithEscape.st
```diff
@@ -5,5 +5,5 @@ testCloseWithEscape
 										keyString: '<escape>';
```
sinnaj-r: Please use pretty print

sinnaj-r (changes requested):
I don't want to be nitpicky but I noticed a few things while testing 😃 :

1. I can't move the cursor with the arrow keys. I understand, that the right arrow key has a different action attached to it, but we might want to change the behaviour so, that it can control the cursor and only apply the "view children"-Action when the cursor is already on the end of the input - or use a different key for the action like 'tab'.
2. When I view the children of an Item and then go back with the left arrow key, the cursor is shown on the left of the query, but when I start typing it will jump to the end of the query.
3. normalKeyStroke: is not tested

packages/Algernon.package/ALGInputField.class/instance/keyboardFocusColor.st
BenBals: This behaivior was adapted from ALGCore. That's there the focus lay before these changes. But feel free to suggest another color.
(This color is chosen not to interfere with the breadcrumbs.)

packages/AlgernonTests.package/ALGTestGeneral.class/instance/testCloseWithEscape.st
BenBals: @robertkirsten could you do this?

BenBals:
> I can't move the cursor with the arrow keys. I understand, that the right arrow key has a different action attached to it, but we might want to change the behaviour so, that it can control the cursor and only apply the "view children"-Action when the cursor is already on the end of the input - or use a different key for the action like 'tab'.

This is also the old behavior. I agree that we should change the keybindings but I think it's its own user story and does not belong to this issue. This was just making the old behavior visible via a cursor. How to you feel about opening up a seperate issue and suggesting it at the next client meeting, @sinnaj-r?

BenBals:
@robertkirsten and @phlprcks do you feel confident in implementing the other two changes @sinnaj-r requested?

Regarding number 2: Take a look at `ALGCore>moveCursorToEnd`. I implemented that method to always move the cursor to the correct position when typing. Now that the cursor is always visible, maybe we should call it earlier?

packages/Algernon.package/ALGInputField.class/instance/keyboardFocusColor.st
sinnaj-r: Okay, all right, I was just wondering what this color has to do with the focus.

sinnaj-r:
> > I can't move the cursor with the arrow keys. I understand, that the right arrow key has a different action attached to it, but we might want to change the behaviour so, that it can control the cursor and only apply the "view children"-Action when the cursor is already on the end of the input - or use a different key for the action like 'tab'.
> 
> This is also the old behavior. I agree that we should change the keybindings but I think it's its own user story and does not belong to this issue. This was just making the old behavior visible via a cursor. How to you feel about opening up a seperate issue and suggesting it at the next client meeting, @sinnaj-r?

That sounds reasonable. I see you've already created an issue. 👍

packages/Algernon.package/ALGInputField.class/instance/keyboardFocusColor.st
BenBals: I think we should add a comment explaining that intent. Could you do that @robertkirsten? (I do not want to create merge conflicts.) I'll write the comment for you:
> Make the keyboard focus color match the background it might cover to hide it.

packages/AlgernonTests.package/ALGTestGeneral.class/instance/testCloseWithEscape.st
robertkirsten: Done.

packages/Algernon.package/ALGInputField.class/instance/keyboardFocusColor.st
robertkirsten: Done.

BenBals:
@robertkirsten do you need help implementing some of the requested changes?

robertkirsten:
I haven't tried yet, I will take a look at it this evening.

kej-jay:
Agree with the previous points. I do not have any other remarks.

kej-jay (dismissed):

kej-jay:
Ich muss gestehen, ich hatte für das Review vorhin auch nur die Tests und coverage laufen lassen, in den Code geguckt und geguckt ob Algernon im Prinzip noch läuft.
Aber was genau da jetzt neu ist und ich nun machen können sollte weiß ich nicht, weil aus der Doku des Issues nicht klar ist, was jetzt wie laufen soll für den User.

Aber afaik läuft es bei mir auch nicht

kej-jay (changes requested):
Läuft bei mir doch, hatte Probleme mit Git, dass ich einen alten Commit erhielt von der Branch. 
Die Ansätze 1 und 2 von Jannis sehe ich genauso, imo ist der jetzige Stand nicht ausreichend, um das Issue zu schließen. 
Zwar wurde als Kriterium festgelegt, dass solch ein Cursor existiert, aber die User Story, auf der das Issue basiert listet dies als Funktion "to make editing my request easier" was mit derzeitiger Funktionalität nicht sonderlich gegeben ist.
Ich helfe gerne, Funktionalität hinzuzufügen, denke aber auch, dass wir alle dies mehr hätten gleich klären sollen und definitiv nochmal besprechen sollen, was gewünscht ist. Aber dafür existieren bereits die entsprechenden Issues afaik.

phlprcks:
> 
> 
> I don't want to be nitpicky but I noticed a few things while testing 😃 :
> 
>     1. I can't move the cursor with the arrow keys. I understand, that the right arrow key has a different action attached to it, but we might want to change the behaviour so, that it can control the cursor and only apply the "view children"-Action when the cursor is already on the end of the input - or use a different key for the action like 'tab'.
> 
>     2. When I view the children of an Item and then go back with the left arrow key, the cursor is shown on the left of the query, but when I start typing it will jump to the end of the query.
> 
>     3. normalKeyStroke: is not tested

2: If you look closely the Cursor is also red in the beginning. The color as well as the position will be set once the first keyStroke is handled. As far as I understand it is problematic to do this beforehand because the initialization of Algernon needs to be finalized. Otherwise we would get errors thrown. Maybe we could catch the special case that there is already input and thus an Algernon instance in the mentioned case. I, however, was not able to determine such cases.

3: I wrote a test that checks the cursors color after a normalKeyStroke. The other Keystrokes are basically only handed to super. Does someone else have an idea on what tests to do there?

sinnaj-r:
> > I don't want to be nitpicky but I noticed a few things while testing 😃 :
> > ```
> > 1. I can't move the cursor with the arrow keys. I understand, that the right arrow key has a different action attached to it, but we might want to change the behaviour so, that it can control the cursor and only apply the "view children"-Action when the cursor is already on the end of the input - or use a different key for the action like 'tab'.
> > 
> > 2. When I view the children of an Item and then go back with the left arrow key, the cursor is shown on the left of the query, but when I start typing it will jump to the end of the query.
> > 
> > 3. normalKeyStroke: is not tested
> > ```
> 
> 2: If you look closely the Cursor is also red in the beginning. The color as well as the position will be set once the first keyStroke is handled. As far as I understand it is problematic to do this beforehand because the initialization of Algernon needs to be finalized. Otherwise we would get errors thrown. Maybe we could catch the special case that there is already input and thus an Algernon instance in the mentioned case. I, however, was not able to determine such cases.
> 
> 3: I wrote a test that checks the cursors color after a normalKeyStroke. The other Keystrokes are basically only handed to super. Does someone else have an idea on what tests to do there?

2: I could also reproduce the error with the cursor color. 
At least for the wrong position I just made a fix-commit. Actually these are only two lines that manually reset the cursor to the end. 
The color problem is not that critical, but you might be able to integrate it at the position of the fix.

3: I would add a check to test if the letter from the event actually appears in the text field (Meaning, if the "contents"-Attribute gets set). But otherwise it's good that the function is now tested at all.

sinnaj-r (approved):
Since my points have now been dealt with except for very minor adjustments, I will now approve the PR. We will have to discuss with the customer how we will deal with the missing features (aka. #73).

Issue #56, opened 2020-05-21
 Actually implement a Fuzzy-Matching Algorithm


sinnaj-r:
Algernon currently does not do fuzzy matching. It does regular substring matching and uses levenshtein distance to sort only after the filtering.

Try to enter `ALGKore`, it should return the `ALGCore` class, but it doesn't.

This might require significant reengineering of internal infrastructure.

# Acceptance Criteria
| Client | Team | Description |
|--|--|--|
| :heavy_check_mark: | :x: | Typing the initials of a query makes the result show up |
| :heavy_check_mark: | :x: | Fuzzy matching works with hotword queries |
| :heavy_check_mark: | :x: | Search does not feel slow |

# Old Version
As a user I want search results that are _closer_ to my search term. 
E.g. when typing "ALG" I want result for the classes "ALG..." and not "ABC".

So we have to replace the Levenshtein distance with a more suitable algorithm.

BenBals:
See https://en.wikipedia.org/wiki/Approximate_string_matching for a basic introduction.

BenBals:
For better metrics than Levenshtein distance see https://en.wikipedia.org/wiki/String_metric.

BenBals:
See https://arxiv.org/pdf/2004.08350.pdf for a comprehensive survey.

BenBals:
One Algorithm I just came up with (that's definitely not original to me) is to do the following:
1. All characters of the search must be in the result in the right order.
2. We score the result by the number of characters between the first and the last of the query.

E. g. `Algro` would match `Algernon` and have a (negative) score of 3 as the letters `e` and `n` (twice) are missing..

This behavior is similar to what a lot of fuzzy matchers do. It's not quite as powerful, but has certain advantages.

- Can be computed in O(n) and with far lower constants.
- Gives better result for short queries or long search strings.

BenBals:
@MrModder we just found out that Algernon does not do fuzzy matching. What do you think?

BenBals:
@kej-jay We should add acceptance criteria. Do you have any ideas? Mine feel inprecise.

Issue #19, opened 2016-07-07
directly search for methods


annkatrinkuessner:
As a developer I want to find methods without specifying the class first in order to reduce the time it takes to find something in Algernon.

# Acceptance Criteria
- [x] algernon finds methods without needing a specified class

# Old Description
Allow for a direct search for methods on top level navigation.
Currently search for methods is only possible within classes (class view when right arrow key was hit).

sinnaj-r:
The code for this is already existing and fully functional. See MethodType algertemList and a lot of method related functions on the ALGItem; the main problem seems to be that speed because of the ~80.000 added items to search through.

BenBals:
Maybe it ties into #1 then?

PR #82, opened 2020-06-04
Directly search for methods, #19

BenBals:
NB. that currently we do not recommend merging this PR. It makes Algernon noticeably slower because we add about 70'000 items to search through. That can only be mitigated by solving #56 and #1.


BenBals:
@MrModder how do you feel about the speed trade-off?

You could take a look at my proposal in #56. If you like that behavior, we could to a spike to see if it mitigates the speed problems.

LeonMatthes:
### Customer feedback
The lag in my image doesn't seem too bad, it's somewhat noticeable, but could be worse.
Further performance optimizations can be handled as User stories (with spikes as necessary).

### Tutor feedback
As developers, taking care to not degrade performance to the point of unusability is your own responsibility, and you are free to make whatever changes necessary. Take care to consider this when estimating a User Stories effort.
Small degradations like these are acceptable, however.

kej-jay (changes requested):
Test ran through, manual testing worked well too and coverage is not lower than the value I knew of. 
Good job you both.

One single thing that is not causing any crashes or so, but if you'd explain it to me if you know why this is happening:
The debug console opens up when running the tests and gives the following errors:
_LoadLibrary(UUIDPlugin) (998: Unzuläÿssiger Zugriff auf einen Speicherbereich.)
LoadLibrary(UUIDPlugin.dll) (998: Unzuläÿssiger Zugriff auf einen Speicherbereich.)
LoadLibrary(SqueakSSL) (998: Unzuläÿssiger Zugriff auf einen Speicherbereich.)
LoadLibrary(SqueakSSL.dll) (998: Unzuläÿssiger Zugriff auf einen Speicherbereich.)_

Again, no tests crashed because of this, but it seems suspicious to me, so I wanted to ask you, if you have a clue what it is about.

BenBals:
I have no idea what that is about. I cannot see how that is related to our changes.

Do you see similar behavior on other branches?

Could you provide the stack trace from the debugger?

BenBals:
@kej-jay have you found out if your issue is related to this pr?

kej-jay:
> 
> 
> @kej-jay have you found out if your issue is related to this pr?

Yes, a clean install stopped it from happening, so I guess my image was causing this.

kej-jay (approved):

robertkirsten (approved):
Looks good from my point of view too, all tests passed, code coverage was increased and searching for methods works very good!

PR #90, opened 2020-06-16
fix #87

### LINKED ISSUE #87: ZeroDivide Error when using arrow keys to navigate Algernon results
Steps to reproduce:

1. start algernon and type "algernontest"
2. use arrow down so that AlgernonTest Category (not package) is highlighted
3. use arrow right and use arrow down so that ALGLinting is highlighted
4. use arrow right to select ALGLinting
5. now use arrow left

sinnaj-r:
Bug is fixed.

Closes #87

### DIFF
--- a/packages/Algernon.package/ALGInputField.class/instance/moveCursorToEnd.st
+++ b/packages/Algernon.package/ALGInputField.class/instance/moveCursorToEnd.st
@@ -1,5 +1,5 @@
 event handling
 moveCursorToEnd
 	self editor clearSelection.
-	self size > 0 ifTrue:[self cursorWrapped: self size].
-	self editor selectFrom: self contents size+1 to: self contents size.
+	self cursorWrapped: self size.
+	self editor selectFrom: self contents size + 1 to: self contents size
--- a/packages/Algernon.package/ALGInputField.class/instance/moveCursorToEnd.st
+++ b/packages/Algernon.package/ALGInputField.class/instance/moveCursorToEnd.st
@@ -1,5 +1,5 @@
 event handling
 moveCursorToEnd
 	self editor clearSelection.
-	self cursorWrapped: self size.
+	self size > 0 ifTrue:[self cursorWrapped: self size].
 	self editor selectFrom: self contents size+1 to: self contents size.
--- a/packages/AlgernonTests.package/ALGTestGeneral.class/instance/testAdvancedChildNavigation.st
+++ b/packages/AlgernonTests.package/ALGTestGeneral.class/instance/testAdvancedChildNavigation.st
@@ -1,0 +1,11 @@
+testing
+testAdvancedChildNavigation
+	| oldQuery |
+	self runQuery: 'AlgernonTests'.
+	self moveCursorTo: 'Category'.
+	oldQuery := self currentQuery.
+	self algerInstance showChildList.
+	self algerInstance showChildList. "Navigate all the way in"
+	self algerInstance showParentList.
+	self algerInstance showParentList.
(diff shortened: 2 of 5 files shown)


packages/Algernon.package/ALGInputField.class/instance/moveCursorToEnd.st
```diff
@@ -1,5 +1,5 @@
 event handling
 moveCursorToEnd
 	self editor clearSelection.
-	self cursorWrapped: self size.
```
BenBals, line 4: I think this behavior should live in cursorWrapped as that method is already concerned with cursor special cases.

sinnaj-r: So should we overwrite the function ? Because currently it's a function only defined on TextMorph.

BenBals: I think that might be a good idea. Sounds like we want to extend the wrapping behavior. What do you think?

packages/AlgernonTests.package/ALGTestGeneral.class/instance/testAdvancedChildNavigation.st
```diff
@@ -0,0 +1,11 @@
...
+testAdvancedChildNavigation
+	| oldQuery |
+	self runQuery: 'AlgernonTests'.
+	self moveCursorTo: 'Category'.
+	oldQuery := self currentQuery.
+	self algerInstance showChildList.
+	self algerInstance showChildList. "Navigate all the way in"
+	self algerInstance showParentList.
+	self algerInstance showParentList.
+	self assert: (oldQuery = (self currentQuery))
```
BenBals, line 11: Why would this trigger the size to be 0?

BenBals (approved):
Looks good to me iG. Feel free to act on my comments as you see fit.

BenBals:
Closes #87

packages/Algernon.package/ALGInputField.class/instance/moveCursorToEnd.st
sinnaj-r: I see your Point. Just moved the logic to an override.

packages/AlgernonTests.package/ALGTestGeneral.class/instance/testAdvancedChildNavigation.st
sinnaj-r: Because the search term is deleted when the children are accessed; thus has the length 0.
(I made the test a bit more general, because we don't have a test for viewing children more than one time)

BenBals: Why is the search term even deleted when we view children? Seems like a weird behavior to me.

BenBals: I think we should discuss this with the client. Could you write an additional, lower-level test that clears the term "manually" to be future proof should this behavior change?

sinnaj-r: Because you should be able to search trough the Children.
(And the original term is restored, when you navigate back to the parent)

BenBals: Ah! Now I see. That's non-obvious. Maybe we should have an indicator.

BenBals (approved):

PR #96, opened 2020-06-16
Fix #95

### LINKED ISSUE #95: "help do" should return the "do"-help-entry as first result

sinnaj-r:
See #95


sinnaj-r:
This is a greater problem than I thought, not only the help filter was faulty, but also the Action Filter.
It's quite fascinating that we didn't test the result order at all...

sinnaj-r:
We also need a test if the expression item is shown. I broke that feature and no test failed.
(just one for the _classic_ "includesSubstring:"-Deprecation)

BenBals:
This should all be fixed with better fuzzy matching.

sinnaj-r:
> This should all be fixed with better fuzzy matching.

Thats right - and we should convince the customer that that issue/feature is important.

But for now my fixes should at least make it work as described. 
The filter:given: code for the Action Filter was completely faulty.

BenBals (approved):

PR #99, opened 2020-06-18
Fuzzy Matching, Implement #56

### LINKED ISSUE #56: Actually implement a Fuzzy-Matching Algorithm

BenBals:


BenBals:
It turns out actually implementing fuzzy matching gets rid of a lot of special casing. We'll have to see if that leads to any funny behavior.

BenBals:
Please run `ALGInstallation instance cleanup` before evaluating this PR as results are cached.

BenBals:
Our changes have all children disappear. Does anyone have an idea why that might happen? (@sinnaj-r?) Take a look at the tests.

sinnaj-r:
> Our changes have all children disappear. Does anyone have an idea why that might happen? (@sinnaj-r?) Take a look at the tests.

The problem is your algo doesn't show result for an empty string. Therefore the children are not gone, but just not shown; if you search e.g. "browse" the related actions does show up.

BenBals:
Thank you. What should the correct return value for the empty string be?

sinnaj-r:
> Thank you. What should the correct return value for the empty string be?

In the old System we had the method relevantForEmptyString. if this returned true for a item it would be shown without a search term. We might need something similar.

BenBals:
Hmm, I just removed all that code...

sinnaj-r:
> Hmm, I just removed all that code...

I saw that :D

sinnaj-r:
> > Hmm, I just removed all that code...
> 
> I saw that :D

Maybe it's enough to just build an exception for ActionItems - but I don't know if there are any side effects / if other items required that option.

BenBals:
@MrModder, what behavior do you want when Algernon has an empty query? Currently nothing is shown, probably for performance reasons.

Now we could show something as an example to look at because we have improved the performance for that case significantly. In this branch when you open Algernon you see nothing, but when you clear it out, you get results for the empty query. We think that is the right behavior.

BenBals:
We feel that the performance is good enough, but that is for the customer to decide.

packages/Algernon.package/ALGItemFilter.class/instance/computeFuzzyItemsOutOf.given..st
```diff
@@ -0,0 +1,9 @@
+filtering
+computeFuzzyItemsOutOf: anItemCollection given: aStringCollection
+	| fuzzyMatcher fuzzySortedItems queryString |
+	aStringCollection first isEmpty ifTrue: [ ^OrderedCollection new ].
+	fuzzyMatcher := ALGFuzzyMatcher new.
```
sinnaj-r: Should we use a class method here ?

packages/AlgernonTests.package/ALGTestFuzzySearch.class/instance/testScore0CaseInsensitive.st
```diff
@@ -0,0 +1,5 @@
+testing
+testScore0CaseInsensitive
+	"Make sure normal input works"
+	self runQuery: 'aLGERNON'.
```
sinnaj-r, line 4: Should we test for the position of the item as well ? (Regards all testScore-Tests)

packages/Algernon.package/ALGActionVerbFilter.class/instance/specialSelection.given..st
```diff
@@ -0,0 +1,6 @@
+accessing
+specialSelection: anItemCollection given: anActionVerb
+	(anActionVerb = '=' or: [anActionVerb = 'do'])
+		ifTrue: [ ^ anItemCollection select: [ :each | each algerType class = ALGExpressionType] ].
```
sinnaj-r: I know, this behaviour origins in the legacy code, but I think we should try to move this special casing to the ExpressionItem. E.g. the Existing "matchesAction"-Method. It would make the code cleaner and easier to implement more "special selections" in the Future.

sinnaj-r (changes requested):
First of all: Great Work! Imho the performance is really good and new algorithm was implemented nicely while removing a lot of the overly complex legacy code 👍 
My comments therefore only address a few minor details that I have noticed.

packages/AlgernonTests.package/ALGTestFuzzySearch.class/instance/testScore0CaseInsensitive.st
BenBals: Theoretically the testScore test cover the ranking, as it is not explicit but derived from that. Also that would be more a test of the SmartSorter as well.

What do you think?

packages/Algernon.package/ALGActionVerbFilter.class/instance/specialSelection.given..st
BenBals: The thing is that `matchesAction` and the scoring are two subsequent filters. That means every item is matched against both filters.

That means special casing overwrites the whole fuzzy matching for performance reasons. I don't see how that decision could be made at the item level.

BenBals:
# TODO
- [x] show action children
- [x] move expression item special casing to own filter class
	- We solved this this in a better way by allowing an item to circumvent fuzzy matching by returning true to `isRelevantForAllStrings`

packages/Algernon.package/ALGActionVerbFilter.class/instance/specialSelection.given..st
BenBals: We have moved/refactored this special casing into the item level. An item can set `isRelevantForAllQueries`.

packages/Algernon.package/ALGItemFilter.class/instance/computeFuzzyItemsOutOf.given..st
BenBals: Yep, did it.

BenBals:
We are quite happy with the performance of the fuzzy matching.

It's so fast, that (at least during our profiling) the class comment search more expensive even though it relies heavily on prebuilt caches.

BenBals:
Should be ready to merge now.

BenBals:
> We feel that the performance is good enough, but that is for the customer to decide.

Currently the performance is better than it was before. If @sinnaj-r thinks the same, I think we should merge this pr. The customer should only decide if the performance is worse.

sinnaj-r (changes requested):
First of all the performance is very good! 
But there is something wrong with searching children. When I search for "morph", then look at the children and search for "step", there is sometimes no result. Other times there were only the methods among the top results, not the actions. (But at least you could search for the actions)

BenBals:
Have you cleaned the caches?

sinnaj-r:
> Have you cleaned the caches?

I did a normal cleanup. Is there anything more I should do ?

BenBals:
>  But there is something wrong with searching children. When I search for "morph", then look at the children and search for "step", there is sometimes no result. Other times there were only the methods among the top results, not the actions. (But at least you could search for the actions)

That happens because Morph has more than 1000 methods.

Because of performance reasons we cut off results for an empty query after 1000.

Do you know a way around that, @sinnaj-r?

BenBals:
I raised the limit, so now it should be fine.

BenBals:
Can you try it out?

packages/Algernon.package/ALGItemProvider.class/instance/evaluateSearchFor..st
```diff
@@ -1,8 +1,8 @@
 filtering
 evaluateSearchFor: aString
-
 	| filterStrategy resultItems tokens |
 	tokens := self tokenize: aString.
 	filterStrategy := self itemFilterFor: tokens.
 	resultItems := filterStrategy filter: self items given: tokens.
-	^ self sorter sort: resultItems given: aString
\ No newline at end of file
+	self halt.
```
sinnaj-r: Please remove the halt :)

BenBals: Of course. I'm sorry. It's gone.

sinnaj-r (approved):
Everything is working for me now. I only noticed that the children of categories don't have the actions in the top results, but it seems to be a legacy problem.

Issue #93, opened 2020-06-16
Algernon keeps position when resizing window


sinnaj-r:
As a user I want Algernon to be opened within my view even if I changed the monitor resolution in order to see Algernon properly.

More info:
When I open Algernon on an ultrawilde monitor and than on a narrow monitor algernon is out of view. We should reset the repositioning when the resolution of squeak changes.

# Acceptance Criteria
| Team | Customer | Title |
|------|----------|-------|
| :x: | :heavy_check_mark: | When I switch to another resolution, a new Algernon opens within my view |

BenBals:
What you are observing is the behavior of all of Squeak.

sinnaj-r:
That's true, but other squeak-windows can be just closed and opened again; getting Algernon to show up again is far more difficult for the user. 
Also I think it would be quite easy to just remember the resolution and reset the location when it was changed since the last Algernon activation.

BenBals:
That's true, but I don't think it's a bug. I think we should estimate it and "bill" the client for it as it is a new feature.

phlprcks:
> 
> 
> That's true, but I don't think it's a bug. I think we should estimate it and "bill" the client for it as it is a new feature.

Same opinion. I edited the description accordingly.

PR #104, opened 2020-06-25
Implement #93

kej-jay:
Please review. 
We implemented a repositioning of Algernon when the Window was made smaller after the last opening of Algernon. It should then be positioned in the middle of the screen again.
When the screen gets bigger(or stays the same), the position of Algernon stays the same.


packages/Algernon.package/ALGCore.class/instance/deactivate..st
```diff
@@ -3,6 +3,7 @@ deactivate: hand
 	self active
 		ifFalse: [^ self].
 	self lastInputString: self topmostQuery.
+	self lastWorldBounds: World bounds corner.
```
sinnaj-r: You need to insert a checkWorldBounds before updating it. Because when I resize the window while Algernon is opened and close Algernon after the resize the new resolution is saved, therefore checkWorldBounds doesn't do anything on the next activation; but Algernon could still be out of view.

packages/Algernon.package/ALGCore.class/instance/checkWorldBounds.st
```diff
@@ -0,0 +1,6 @@
+visibility
+checkWorldBounds
+
+	(self lastWorldBounds x > World bounds corner x
```
sinnaj-r: This doesn't check for the current Position of Algernon at all. If I remember correctly, we wanted to preserve the Position of Algernon if it is still in View.

In this context it may also be reasonable to check if the Algernon bounds are negative, that is, if the window was moved left out of the image.

packages/AlgernonTests.package/ALGTestUI.class/instance/testAlgernonPositioningWindowIsSameSizeAfterReopening.st
```diff
@@ -0,0 +1,8 @@
+testing
+testAlgernonPositioningWindowIsSameSizeAfterReopening
+	| point oldPosition|
+	point := 0 @ 0.
```
sinnaj-r: You can remove the "point" here.

packages/AlgernonTests.package/ALGTestUI.class/instance/testAlgernonPositioningWindowIsSmallerAfterReopening.st
```diff
@@ -0,0 +1,7 @@
+testing
+testAlgernonPositioningWindowIsSmallerAfterReopening
+	| point |
+	point := 1 @ 1.
+	self algerInstance lastWorldBounds: World bounds corner + point.
+	self algerInstance activate.
+	self assert: ((self algerInstance center x = self algerInstance class defaultPosition x rounded) & (self algerInstance center y = self algerInstance class defaultPosition y rounded))
```
sinnaj-r: You don't move the Algernon Instance in this Test, therefore this will always succeed.

sinnaj-r (changes requested):

packages/AlgernonTests.package/ALGTestUI.class/instance/testAlgernonPositioningWindowIsSameSizeAfterReopening.st
kej-jay: We had problems adding the point without having it saved in a variable.

packages/AlgernonTests.package/ALGTestUI.class/instance/testAlgernonPositioningWindowIsSmallerAfterReopening.st
kej-jay: Algernon is repositioned through the check between lastWorldBounds and the current world bounds when opened. Therefore it will reposition

packages/Algernon.package/ALGCore.class/instance/checkWorldBounds.st
kej-jay: As far as we know, we wanted to move algernon to the middle of the Screen when the screen has been made smaller and keep the position when the window gets bigger

packages/Algernon.package/ALGCore.class/instance/deactivate..st
kej-jay: I think our tutor even said that its okay to have the repositioning only affect Algernon after closing and reopening, so it is a valid process as far as I know
Or maybe I do not understand your point very well

sinnaj-r: Yes that's correct, but when the resizing is done while Algernon is open (meaning after the activation and call of checkWorldBounds and before the deactivation and the setting of lastWorldBounds) your code saves the new window size on closing Algernon, and when Algernon is activated again checkWorldBounds does nothing, even trough Algernon could be out of view.

To try it for your self, just open Algernon, drag it on the right edge, keep it open and make the squeak window smaller, when you now reopen Algernon it will be still out of view.

packages/Algernon.package/ALGCore.class/instance/checkWorldBounds.st
sinnaj-r: I'm not sure. @BenBals what do you think? Should we check if the Algernon is out of view due to the scaling down of the window or just recenter it no mater what ?

packages/AlgernonTests.package/ALGTestUI.class/instance/testAlgernonPositioningWindowIsSameSizeAfterReopening.st
sinnaj-r: Right, but adding a point with the value 0@0 won't do that much. 😃 You can remove the point variable and summation altogether.

packages/AlgernonTests.package/ALGTestUI.class/instance/testAlgernonPositioningWindowIsSmallerAfterReopening.st
sinnaj-r: Algernon is in center of the World at the start of the test; If your methods run it would be repositioned in the center of the world. If your methods don't run/do nothing it would still be in the center of the world. Therefore the tests will allways succeed, wether your logic is working or not.
(To verify just comment out the "activate" statement, which runs your logic; the tests will work nevertheless)

packages/Algernon.package/ALGCore.class/instance/checkWorldBounds.st
BenBals: We should only move Algernon if required.

BenBals: If I'm correct, that would also mean that we don't need a instance variable, making the computation stateless, which should be preferred.

packages/Algernon.package/ALGCore.class/instance/deactivate..st
BenBals: That issue would probably also be resolved by just checking if Algernon is in view, as suggested in my other comment.

packages/AlgernonTests.package/ALGTestUI.class/instance/testAlgernonPositioningWindowIsSmallerAfterReopening.st
```diff
@@ -0,0 +1,7 @@
+testing
```
BenBals, line 1: Please also add tests for the case that the window resized but Algernon is still in view, as discussed above.

BenBals (changes requested):
Looks good to me iG. I'm positively surprised that you managed to implement this feature in so little code. The code itself is of really good quality.

Just implement the behavior discussed in the comments and this seems good to go.

packages/AlgernonTests.package/ALGTestUI.class/instance/testAlgernonPositioningWindowIsSameSizeAfterReopening.st
```diff
@@ -0,0 +1,8 @@
...
+	oldPosition := self algerInstance position.
+	self algerInstance lastWorldBounds: World bounds corner - point.
```
BenBals: Please look at the errors squeak gives you. They will tell you what has done wrong. If you did that, you'd easily see that the errors are because you removed `lastWorldBounds` from ALGCore, but still call it from the tests.

BenBals (changes requested):
The changes look good, but the tests fail due to a trivial error. Please fix it and then **write actual tests** changing the world bounds.

BenBals:
I have written you a basis for the kinds of tests you need.

BenBals (approved):
I've now done the needed test changes myself, so we have it ready for the customer meeting.

BenBals:
@robertkirsten @kej-jay Please fix the conflicts by merging development into this branch. (It's important that you do it that way round.)

packages/AlgernonTests.package/ALGTestUI.class/instance/testAlgernonPositioningWindowIsSameSizeAfterReopening.st
```diff
@@ -0,0 +1,6 @@
+testing
+testAlgernonPositioningWindowIsSameSizeAfterReopening
```
sinnaj-r, line 2: Maybe we should move Algernon beforehand ?

packages/AlgernonTests.package/ALGTestUI.class/instance/testAlgernonPositioningWindowIsSmallerAfterReopening.st
```diff
@@ -0,0 +1,9 @@
...
+testAlgernonPositioningWindowIsSmallerAfterReopening
+	| oldExtent |
+	oldExtent := World extent.
+	World extent: (100@100).
+	self algerInstance deactivate.
+	self algerInstance activate.
+	self assert: ((self algerInstance center x = self algerInstance class defaultPosition x rounded) & (self algerInstance center y = self algerInstance class defaultPosition y rounded)).
```
sinnaj-r: You can just write `self algerInstance center =  self algerInstance class defaultPosition rounded`

packages/Algernon.package/ALGCore.class/instance/checkWorldBounds.st
```diff
@@ -0,0 +1,6 @@
+visibility
+checkWorldBounds
```
sinnaj-r, line 2: This is an Edgecase, but we might still want to check if our bounds are smaller than the Worldbounds meaning negative; E.g. Algernon was moved out of view to the top/left instead of the bottom/right.

sinnaj-r (approved):
I generally approve the changes now. 
If you have time before the meeting it would be great if you could look into my comments as well, but otherwise just merge it.

packages/Algernon.package/ALGCore.class/instance/checkWorldBounds.st
BenBals: I don't agree. I don't see why you should do that, but that can only happen if the user deliberately does it. We should not overrule the user.

packages/AlgernonTests.package/ALGTestUI.class/instance/testAlgernonPositioningWindowIsSmallerAfterReopening.st
BenBals: Did that.

packages/AlgernonTests.package/ALGTestUI.class/instance/testAlgernonPositioningWindowIsSameSizeAfterReopening.st
BenBals: Did that.

Issue #103, opened 2020-06-25
Reuse cache in computing results


BenBals:
The new fuzzy matching algorithm has a nice "transitive" (kinda, not sure what the math word really is) property.

If a string matches a pattern then it must also match all substrings of that pattern. We can use that to make Algernon much faster:

Example you search for "Test String". Depending on your typing speed Algernon will compute results at multiple times during that query: Importantly the image freezes during that time. Assume, for illustration, that we compute after "Test" and after the full string "Test String".

Currently we filter all the items against "Test" and then filter all the items against "Test String".

But what we could do, is filter the result of "Test" against "Test String" to obtain the final result. The best thing is that we currently cache the result of "Test" (of all queries, really).

We could even use the entire cache for this kind of optimization easily, not just results from the current query. (Honestly: It would be harder *not* not to use the whole cache.)

That way we can improve the performance of Algernon considerably, especially if you often search for similar strings, e.g ("ALGCore" and "ALGItemProvider" searches could both use the cache for "ALG").

That way we could really increase the use and usefulness of the cache considerably. And the best thing is: I think we could do that in surprisingly little code.

# Acceptance Criteria
| Team | Client | Description |
|-|-|-|
| :heavy_check_mark: | :heavy_check_mark: | Search feels faster. | 
| :heavy_check_mark: | :heavy_check_mark: | Repeating a query feels instant. |
| :heavy_check_mark: | :heavy_check_mark: | Searching for `ALGCore` after searching for `ALGFakeTest` is faster than searching for `Morph` |

PR #108, opened 2020-06-30
Reuse Cache in Computing Results, Implement #103

### LINKED ISSUE #110: Help Keyword doesn't list all available Help Entries.
When I enter "help" it does not show all Help keywords.
If I set `isRelevantForAllStrings` for the HelpType to return true it works, but than the fuzzy matching for e.g. "help do" is broken. I'm no expert with the new FuzzyMatcher, but I think this should be an easy fix. 😃 

(We should also add a test for this behaviour)

BenBals:
Note that if you don't use metacello for upgrading, you have to run `ALGItemProvider allInstances do: [:each | each cache: ALGCache new]` manually.


packages/Algernon.package/ALGCache.class/instance/getAt..st
```diff
@@ -1,6 +1,6 @@
 caching
 getAt: aCacheKey
```
sinnaj-r: I think we should overwrite the directory methods directly. Having both " at:" and "getAt:" at the same time is not really pretty and might confuse later developers.

packages/Algernon.package/ALGCache.class/instance/putItems.at..st
```diff
@@ -1,4 +1,4 @@
 caching
 putItems: anItemCollection at: aCacheKey
```
sinnaj-r: See comment for "getAt:"; I think this should be "at:put:" (or at least, "at:putItems:")

packages/Algernon.package/ALGCache.class/instance/clear.st
```diff
@@ -1,4 +1,3 @@
 caching
 clear
```
sinnaj-r, line 2: Do we need this function ? As this ist just an "rename" for removeAll, we might want to call removeAll instead of clear.

sinnaj-r (changes requested):
First of all the search is now (in comparison) ridiculously fast. This is an astonishing improvement! Great Job 👍 Perhaps we can also "afford" decrease the filteringThreshold, so that the search runs more often therefore showing the results quicker/more responsive as you type.

**But**: I have the Problem, that certain results are missing. E.g. when I search for Algernon I get no hits and when I search for "ALG" some results (The Tests, the ALGBackground-Class, etc.) are not in the results. I would guess that this is an Cache Reset/Init-Problem, do you have an Idea what I could try ?

packages/Algernon.package/ALGCache.class/instance/clear.st
BenBals: This is because we need to ensure that old calls from `ALGInstallation instance cleanup` need work even if the new code has been loaded. That might happen during upgrade via Montecello. (I think, that stuff is black magic.)

BenBals:
Have you done `ALGItemProvider allInstances do: [:each | each cache: ALGCache new]`?

sinnaj-r:
> Have you done `ALGItemProvider allInstances do: [:each | each cache: ALGCache new]`?

Yes. I also did the full "cleanup" procedure.

packages/Algernon.package/ALGCache.class/instance/clear.st
sinnaj-r: Okay, I see.

packages/Algernon.package/ALGCache.class/instance/getAt..st
BenBals: Did that.

BenBals:
Hmm, I can reproduce that. We will have to look into that. Seems like more than your cleanup.

BenBals:
Oh no. I think I know what this is... This might ruin all the positive effects of this PR. The cache does not do what we think it does:

It stores final results. That means if I search for 'ALG' I might get 100 results, but only 10 are displayed. Then the cache saves those 10 results, not the full list. That means if I search for something that was not in the initial list, it will not show up.

BenBals:
That also explains why everthing is so fast...

BenBals:
The bug should be fixed now. Plase re-review.

sinnaj-r:
> The bug should be fixed now. Plase re-review.

Indeed, the error of missing results is now fixed. I still find the speed very good, especially when calling up cached entries. As I said before, I think it makes sense to reduce the filteringThreshold. 

But unfortunately I noticed a few problems after all: 
- If I enter only one character in the search I have a long loading time. Are such search queries not cached? 
- When using filters there is an error (see the errors in the CI)
- The whole children-search-system seems to be broken, I can find all methods in every children-view

BenBals:
1. Such queries are cached, but as explained in my earlier replies, we now have to cache all results, not just the top ten as previously. 
Additionally, the fuzzy score on each item is overwritten on each query, so we have to recompute it for items in the cache, but that's not the main bottleneck. The thing that's really slow is the sorting.
To solve this we could easily have a second cache that only stores the top ten, that would only be used if the full query matches. What do you think?

2. The CI errors should be easy enough to fix, I'll do it when I'm back at my dev machine.

3. I'll have to take a look at that before I can provide more info.

BenBals:
In general I'm not fully happy with the archtecture around how the Provider, Filter and Cache work with the whole caching thing. If you have any idea how to make better abstractions, feel free to make suggestions.

To clarify, in general I'm happy with what each class does, but I'm not so happy with how the code is structured in the ItemProvider mainly as there is a lot of weird, subtle control flow that ensures that every edge case is met.

BenBals:
> The whole children-search-system seems to be broken, I can find all methods in every children-view.

I cannot reproduce that behavior. Have to cleared all caches using `ALGItemProvider allInstances do: [:each | each cache: ALGCache new]` between your reviews?

sinnaj-r:
>   To solve this we could easily have a second cache that only stores the top ten, that would only be used if the full query matches. What do you think?

Another Cache is not really _smooth_, but if it's the only viable solution we should do it - the performance drop when entering only one character is really annoying.
Speaking of reworks - Why do we need the sorting aka re-ranking each time (for cached results)? If It's because of the frequency/recency feature, we should think about how we could just re-sort only the the - whatever - 100 best results

>   In general I'm not fully happy with the archtecture around how the Provider, Filter and Cache work with the whole caching thing. If you have any idea how to make better abstractions, feel free to make suggestions.

I just tried to reconstruct the exact "data" flow, but I noticed that the method names are not good. We should (maybe in a separate issue) rename "updateItemListWith:", "resultsFor:", "relevantItemsFor:" and "evaluateSearchFor:"; they sound like they all do the same thing and a meaningful order of execution is not clear. Then it might be useful to really define one function where alle the calls are made and the data is clearly moved trough the stages - instead of the somewhat recursive schema now.
e.g. in somewhat pseudo-code:
```
self checkForCache ifTrue:[^ cachedResult].
searchResults := self getAllItems.
searchResults := self refineResultsForFilter:...
searchResults := self refineResultsForSearchTerm:...
self sortResults: searchResults.
self writeToCache: searchResults.
^ searchResults
```

As I was writing this I thought we might want to create a new class for a single Search. There we could store searchTerm, tokenization, etc. as instance variables and hide the complex arguments and process of the search behind simple argument-free methods. What do you think?

sinnaj-r:
> > The whole children-search-system seems to be broken, I can find all methods in every children-view.
> 
> I cannot reproduce that behavior. Have to cleared all caches using `ALGItemProvider allInstances do: [:each | each cache: ALGCache new]` between your reviews?

Ah, no, sorry - I did that and now it's working as supposed! 👍

packages/Algernon.package/ALGItemFilter.class/instance/computeFuzzyItemsOutOf.given..st
```diff
@@ -5,4 +5,4 @@ computeFuzzyItemsOutOf: anItemCollection given: aStringCollection
 	queryString := aStringCollection joinSeparatedBy: ' '.
 	anItemCollection do: [ :item | item fuzzyRank: (ALGFuzzyMatcher computeScore: item given: queryString)].
 	fuzzySortedItems := (anItemCollection select: [ :each | each fuzzyRank isNil not ]) sort: [ :a :b | (b fuzzyRank) < (a fuzzyRank)].
```
sinnaj-r: This line should be removed because it was moved to it's own method - correct ?

BenBals:
Actually the stuff is not that complicated to really warrant its own class. But the worse problem is that ItemProvider in this case functions mainly as a data provider to other Algorithms. Extracting this logic would create a class that in envious of the data that the item provider holds.

I tried renaming and moving a bit of code. Tell me if that helps.

sinnaj-r:
> Actually the stuff is not that complicated to really warrant its own class. But the worse problem is that ItemProvider in this case functions mainly as a data provider to other Algorithms. Extracting this logic would create a class that in envious of the data that the item provider holds.
> 
> I tried renaming and moving a bit of code. Tell me if that helps.

Yes, that helped a bit. At least the function names are no longer completely synonymous. 
But tbh I still don't find it intuitive, but more refactoring is probably going beyond the scope of this issue.

BenBals:
I have done a message tally. It seems that the main bottle neck of single letter searches is the class comment search (or more specifically its quadratic runtime). We could build an additional cache, but that would only be a band aid.

The reason that came up now is that previously we limited the number of results before adding the class comment results. The thing about the class comment results is that they use `addIfNotPresent`, but that's a linear operation **for each item** added. Now that we limit the number of items later that came out to bite us.

In general the whole class comment search stuff is fucked up.Do you have any idea how we could sort that out? Or at least how we could avoid `addIfNotPresent`? Do you know if a set could work? Can we efficently convert it to an OrderedCollection for sorting later on?

```
 - 15770 tallies, 15883 msec.

**Tree**
--------------------------------
Process: other processes
--------------------------------
64.8% {10296ms} [] ProcessorScheduler class>>installIdleProcess
  |64.8% {10296ms} ProcessorScheduler class>>idleProcess
1.8% {281ms} [] EventSensor>>installEventTickler
  1.8% {281ms} EventSensor>>eventTickler
    1.8% {281ms} BlockClosure>>on:do:
      1.8% {281ms} [] EventSensor>>eventTickler
        1.8% {281ms} Delay>>wait
          1.8% {281ms} BlockClosure>>ifCurtailed:
            1.8% {281ms} [] Delay>>wait
--------------------------------
Process: (40) 40332: nil
--------------------------------
33.3% {5292ms} MessageTally class>>spyOn:reportOtherProcesses:
  33.3% {5292ms} MessageTally>>spyEvery:on:
    33.3% {5292ms} BlockClosure>>ensure:
      33.3% {5292ms} [] TheWorldMainDockingBar>>startMessageTally
        33.3% {5292ms} PasteUpMorph>>doOneCycle
          33.3% {5282ms} AnimWorldState(WorldState)>>doOneCycleFor:
            25.2% {4000ms} AnimWorldState(WorldState)>>doOneCycleNowFor:
              |20.5% {3252ms} PasteUpMorph>>runStepMethods
              |  |20.5% {3252ms} AnimWorldState(WorldState)>>runStepMethodsIn:
              |  |  20.5% {3249ms} AnimWorldState(WorldState)>>runLocalStepMethodsIn:
              |  |    20.4% {3247ms} StepMessage(MorphicAlarm)>>value:
              |  |      19.6% {3117ms} ClockMorph(Morph)>>stepAt:
              |  |        19.6% {3116ms} Algernon(ALGCore)>>step
              |  |          19.6% {3116ms} CursorWithMask(Cursor)>>showWhile:
              |  |            19.6% {3116ms} BlockClosure>>ensure:
              |  |              19.6% {3116ms} [] Algernon(ALGCore)>>step
              |  |                19.6% {3116ms} Algernon(ALGCore)>>updateItemList
              |  |                  19.6% {3116ms} Algernon(ALGCore)>>updateItemListWith:
              |  |                    19.6% {3106ms} ALGItemProvider>>resultsFor:
              |  |                      19.5% {3103ms} ALGItemProvider>>searchForItems:
              |  |                        19.5% {3103ms} ALGItemProvider>>filteredItemsFor:
              |  |                          19.5% {3103ms} ALGRegularFilter>>filter:given:
              |  |                            18.1% {2880ms} OrderedCollection>>do:
              |  |                              |18.1% {2878ms} [] ALGRegularFilter>>filter:given:
              |  |                              |  18.1% {2876ms} OrderedCollection(Collection)>>addIfNotPresent:
              |  |                              |    18.1% {2876ms} OrderedCollection(SequenceableCollection)>>includes:
              |  |                              |      18.1% {2876ms} OrderedCollection(SequenceableCollection)>>indexOf:
              |  |                              |        18.1% {2876ms} OrderedCollection>>indexOf:startingAt:
              |  |                            1.1% {168ms} ALGRegularFilter(ALGItemFilter)>>filter:given:
              |  |                              1.0% {166ms} ALGRegularFilter(ALGItemFilter)>>computeFuzzyItemsOutOf:given:
              |4.6% {734ms} AnimWorldState(WorldState)>>displayWorldSafely:
              |  4.6% {734ms} BlockClosure>>on:do:
              |    4.6% {734ms} [] AnimWorldState(WorldState)>>displayWorldSafely:
              |      4.6% {734ms} PasteUpMorph>>displayWorld
              |        4.6% {734ms} PasteUpMorph>>privateOuterDisplayWorld
              |          4.6% {734ms} AnimWorldState(WorldState)>>displayWorld:submorphs:
              |            3.8% {601ms} AnimWorldState(WorldState)>>drawWorld:submorphs:invalidAreasOn:
              |              3.8% {599ms} Array(SequenceableCollection)>>do:
              |                3.8% {599ms} [] AnimWorldState(WorldState)>>drawWorld:submorphs:invalidAreasOn:
              |                  3.8% {599ms} Rectangle>>allAreasOutsideList:do:
              |                    3.8% {599ms} Rectangle>>allAreasOutsideList:startingAt:do:
              |                      3.5% {562ms} [[]] AnimWorldState(WorldState)>>drawWorld:submorphs:invalidAreasOn:
              |                        2.7% {424ms} AnimFormCanvas>>fullDrawMorph:
              |                          2.7% {424ms} PluggableSystemWindow(Morph)>>fullDrawAnimatedOn:
              |                            2.7% {422ms} PluggableSystemWindow(Morph)>>fullDrawOn:
              |                              2.4% {380ms} PluggableSystemWindow(Morph)>>drawSubmorphsOn:
              |                                1.8% {293ms} [] PluggableSystemWindow(Morph)>>drawSubmorphsOn:
              |                                  1.8% {293ms} Array(SequenceableCollection)>>reverseDo:
              |                                    1.8% {293ms} [[]] PluggableSystemWindow(Morph)>>drawSubmorphsOn:
              |                                      1.8% {293ms} AnimFormCanvas>>fullDrawMorph:
              |                                        1.8% {293ms} PluggablePanelMorph(Morph)>>fullDrawAnimatedOn:
              |                                          1.8% {285ms} PluggablePanelMorph(Morph)>>fullDrawOn:
              |                                            1.7% {268ms} PluggablePanelMorph(Morph)>>drawSubmorphsOn:
              |                                              1.6% {257ms} [] PluggablePanelMorph(Morph)>>drawSubmorphsOn:
              |                                                1.6% {257ms} Array(SequenceableCollection)>>reverseDo:
              |                                                  1.6% {257ms} [[]] PluggablePanelMorph(Morph)>>drawSubmorphsOn:
              |                                                    1.6% {257ms} AnimFormCanvas>>fullDrawMorph:
              |                                                      1.6% {257ms} PluggableListMorphByItemPlus(Morph)>>fullDrawAnimatedOn:
              |                                                        1.6% {257ms} PluggableListMorphByItemPlus(Morph)>>fullDrawOn:
            8.1% {1281ms} AnimWorldState(WorldState)>>interCyclePause:
              8.0% {1278ms} Delay>>wait
                7.6% {1211ms} BlockClosure>>ifCurtailed:
                  7.6% {1211ms} [] Delay>>wait

**Leaves**
64.9% {10307ms} ProcessorScheduler class>>idleProcess
18.1% {2876ms} OrderedCollection>>indexOf:startingAt:
9.4% {1492ms} [] Delay>>wait

**Memory**
	old			+0 bytes
	young		-444,256 bytes
	used		-444,256 bytes
	free		+444,256 bytes

**GCs**
	full			0 totalling 0 ms (0% uptime)
	incr			30 totalling 70 ms (0.4% uptime), avg 2.3 ms
	tenures		10,863 (avg 0 GCs/tenure)
	root table	0 overflows

```

sinnaj-r:
In my opinion we should only search for class comments when the comment-search keyword is provided. As a user I would not expect class comment hits in the normal results.

BenBals:
That would be a solution, but we would have to get approval by the customer because that changes the behavior.

BenBals:
I'll try using a set first.

sinnaj-r:
> I'll try using a set first.

Yes, do that. If I have correctly seen this, the set uses a slightly more intelligent method to determine if an element is present.

BenBals:
Oh wow, that was a two line change, but the performance is so much better. It's still there, but barely noticable. See if that's enough for you.

If not I could use Set more consistently. Right now I turn the collection into a set before adding the class comments and convert it to an OrderedCollection afterwards. Maybe we could save some of these actions, by generally using Sets up until the last minute.

BenBals:
> Yes, do that. If I have correctly seen this, the set uses a slightly more intelligent method to determine if an element is present.

Sets should (hopefully, I have not checked, that's just the way to do it) be implemented via hashing and therefore provide an (amortized) O(1) add operation.

BenBals:
Okay, now it's fast, but wrong...

sinnaj-r:
> > > The whole children-search-system seems to be broken, I can find all methods in every children-view.
> > 
> > 
> > I cannot reproduce that behavior. Have to cleared all caches using `ALGItemProvider allInstances do: [:each | each cache: ALGCache new]` between your reviews?
> 
> Ah, no, sorry - I did that and now it's working as supposed! 👍

And I got the same problem again. Steps to reproduce: Run the Cleanup/Reset/New Cache stuff. Search for step, Search for ALGCore, in the Children-View of ALGCore search for "step2" - You will see the result, but it shouldn't be there. 😃

sinnaj-r:
> Oh wow, that was a two line change, but the performance is so much better. It's still there, but barely noticable. See if that's enough for you.
> 
> If not I could use Set more consistently. Right now I turn the collection into a set before adding the class comments and convert it to an OrderedCollection afterwards. Maybe we could save some of these actions, by generally using Sets up until the last minute.

I think it's fine. We could create a separate issue to describe the possible (minor) performance tweaks. 
(Which one might do while polishing for submission.)

BenBals:
> And I got the same problem again. Steps to reproduce: Run the Cleanup/Reset/New Cache stuff. Search for step, Search for ALGCore, in the Children-View of ALGCore search for "step2" - You will see the result, but it shouldn't be there. smiley

I cannot reproduce that. Have you applied the latest patch?

sinnaj-r:
> > And I got the same problem again. Steps to reproduce: Run the Cleanup/Reset/New Cache stuff. Search for step, Search for ALGCore, in the Children-View of ALGCore search for "step2" - You will see the result, but it shouldn't be there. smiley
> 
> I cannot reproduce that. Have you applied the latest patch?

I just downloaded a fresh version of Squeak and Algernon aka. your Branch; The Issue persists. (See attached GIF)
Jul-01-2020 23-48-26

sinnaj-r:
> > > And I got the same problem again. Steps to reproduce: Run the Cleanup/Reset/New Cache stuff. Search for step, Search for ALGCore, in the Children-View of ALGCore search for "step2" - You will see the result, but it shouldn't be there. smiley
> > 
> > 
> > I cannot reproduce that. Have you applied the latest patch?
> 
> I just downloaded a fresh version of Squeak and Algernon aka. your Branch; The Issue persists. (See attached GIF)

What about you, @robertkirsten, can you reproduce the issue ?

BenBals:
Thanks for the GIF. That really helped. The issue should be fixed now. (See the commit message for an explanation.)

BenBals:
What's your opinion on the second cache? Should we implement it?

sinnaj-r:
> What's your opinion on the second cache? Should we implement it?

That depends, strictly speaking it's outside the scope, but I have to admit that I personally find the speed boost very appealing. So in the end, you will have to estimate how long it could take and if you want to put in the extra time. As you correctly pointed out today, nothing is really _trivial_ when it comes to the cache.

sinnaj-r (approved):

robertkirsten (approved):

BenBals:
If it is significant for the performance, I don't consider it out of scope, because we were tasked to improve cached performance, not worsen it (even if only in certain cases). Therefore I implemented the feature. Please review the new code. After that I consider the pr ready to merge. (If the tests pass.)

sinnaj-r (approved):

Issue #34, opened 2020-05-03
Hide top results from full list


BenBalsInsecure:
As a user I want top results to be hidden from the full result list because I'm confused by seeing the same result twice and want to better utilize the screen real estate.

# Acceptance Criteria
| Team | Client | Description |
|-|-|-|
| :heavy_check_mark:  | :heavy_check_mark:  | any result that is displayed in the top results is not displayed again in the full list |

LeonMatthes:
Estimated effort: 5

PR #109, opened 2020-07-01
Fix #98 #78 #34

### LINKED ISSUE #98: Algernon can not be dragged and dropped directly after opening it again

sinnaj-r:
This PR should prove quick fixes for #98 and #78.
Also #34 should be implemented.


sinnaj-r:
The fix for #78 is not that great, but it's in line with the other cache invalidation that Algernon does. If we want a more elaborate Invalidation, we should create a separate ("_billable_") user story.

sinnaj-r:
The main fixes/implementations are done now - so I'm removing the draft flag; But I didn't run the coverage-check yet, therefore some one or two additional tests may be necessary.

packages/Algernon.package/ALGTopCategorizer.class/instance/categorize.withExistingCategories..st
```diff
@@ -0,0 +1,9 @@
+categorizing
+categorize: anItemList withExistingCategories: aCategoriesCollection
```
BenBals, line 2: Is there a reason why `aCategoriesCollection` is not used in this method?

BenBals (approved):
Looks good to me.

packages/Algernon.package/ALGTopCategorizer.class/instance/categorize.withExistingCategories..st
sinnaj-r: We just don't need it, as the TopCategories are generated independently from other categorizers.

I could have overwritten the `categorize:` Method, but then I would have to call it from the `categorize:withExistingCategories:` Method, which would invert the "flow" of the super class. Therefore I thought it would be acceptable to just have an unused argument.

BenBals: Yeah, I thought that, but wanted to check in.

PR #112, opened 2020-07-02
Fix #111: Tests are too slow

### LINKED ISSUE #111: Tests are too slow

sinnaj-r:
Please merge after #108.


sinnaj-r:
@BenBals When I refactored the calls of `self algerInstance` to a class call as discussed, I noticed that this makes the tests on average noticeable longer (`self algerInstance <-> self class algernonInstance`). 
So I renamed `self algerInstance` anyway, but only to `self algInstance` and pointed out in the method comment that this is only a shortcut. What do you think, is this okay, or should I change it to a class call nevertheless ?

Apart from that, the PR is now ready for review - and afterwards ready for merging.

BenBals:
I think thats fine.

BenBals (approved):
The test suite is not fast by any means, it takes just over a minute on my machine, but this PR made them actually runnable. Good work.

PR #116, opened 2020-07-06
Fix bug with #111

sinnaj-r:


BenBals:
Looks good. Why does this fix the issue?

BenBals (approved):

sinnaj-r:
> Looks good. Why does this fix the issue?

Before, only the timestamp of the current day was retrieved. So the difference was always 0, now I use the correct timestamp function.

PR #121, opened 2020-07-14
rewrite cleanup

BenBals:
# Please do the following to test before you approve.

1. Check out master.
1. Use Algernon
1. Run the upgrade snippet below. (Will be once the pr is ready for review.)
1. Test that Algernon still works.


BenBals:
@sinnaj-r The tests that fail in ci pass on my machine. Could you try them out?

Possible causes/notes
- The ExpressionItems have to know the controller (ALGCore) at creation
- Therefore the ItemProvider has to know the ALGCore before creating items
- That's ensured by using `ItemProvider>>newIn:` in `ALGCore>>initializeProvider`

That should work, but in the CI the controller is not set on the expression items (it'ss nil).

sinnaj-r:
> @sinnaj-r The tests that fail in ci pass on my machine. Could you try them out?
> 
> Possible causes/notes
> 
> * The ExpressionItems have to know the controller (ALGCore) at creation
> * Therefore the ItemProvider has to know the ALGCore before creating items
> * That's ensured by using `ItemProvider>>newIn:` in `ALGCore>>initializeProvider`
> 
> That should work, but in the CI the controller is not set on the expression items (it'ss nil).

It's running on my machine as well. 
I can just guess, but I think it would be possible that `ALGItemProvider>cleanUp` was called. (As cleanUp is an "system"-method)

BenBals:
Hmm, renaming to `cleanup` does not help.

BenBals:
Squeak decided not to push the code anymore, but to overwrite the upstream branch incorrectly every time. See the new pr.

BenBals:
See #125 .

Issue #84, opened 2020-06-11
ResultItems are clickable


kej-jay:
As a User I want to be able to click on the single result items.

# Acceptance Criteria
| Team | Customer | Title |
|------|----------|-------|
| :heavy_check_mark: | :heavy_check_mark: | Clicking on a result item does not close Algernon |
| :heavy_check_mark: | :heavy_check_mark: | Clicking on a result item performs the associated action |
| :heavy_check_mark: | :heavy_check_mark: | Hovering over a result item gives a visual indication |

Issue #57, opened 2020-05-21
Unify searching for class comments


BenBals:
As a new user I want a single, easy to remember hotword to search for class comments.

Currently we have ?, which and apropos.

PR #128, opened 2020-07-20
Implement #84 & #57

sinnaj-r:


packages/Algernon.package/ALGCore.class/instance/mouseDown..st
```diff
@@ -0,0 +1,5 @@
+initialization
+mouseDown: evt
```
BenBals, line 2: Is there a reason the ResultItems or the ResultList can't handle mousedown themselves?

packages/Algernon.package/ALGCore.class/instance/mouseMove..st
```diff
@@ -0,0 +1,5 @@
+initialization
```
BenBals: Is this really initialization?

packages/Algernon.package/ALGCore.class/instance/handlesMouseMove..st
```diff
@@ -0,0 +1,3 @@
+initialization
+handlesMouseMove: evt
```
BenBals, line 2: Why do we need to handle mouse move? Can't we use mouse hover?

packages/Algernon.package/ALGResultList.class/instance/handlesMouseDown..st
```diff
@@ -1,4 +1,4 @@
 event handling
 handlesMouseDown: evt
```
BenBals: Do we need to explicitly set this to false? Can't the method be removed.

packages/Algernon.package/ALGResultList.class/instance/listItemAtPosition..st
```diff
@@ -0,0 +1,13 @@
+drawing
+listItemAtPosition: aPoint
```
BenBals, line 2: Do we really need to do this?

packages/Algernon.package/ALGResultList.class/instance/mouseMove..st
```diff
@@ -0,0 +1,5 @@
+event handling
+mouseMove: evt
+	| newSelection |
+	newSelection:= (self listItemAtPosition: evt position).
```
BenBals: You don't need the temporary variable here.

packages/AlgernonTests.package/ALGTestActionVerbs.class/instance/testHelpItemsQuestionMark.st
```diff
@@ -0,0 +1,5 @@
+testing
+testHelpItemsQuestionMark
+	self runQuery: '?'.
+	self assert: (self items allSatisfy: [:each | each algerType typeName = 'Help' ]).
```
BenBals: This seems duplicated. Maybe move to helper?

packages/AlgernonTests.package/ALGTestUI.class/instance/testClickOnResultItem.st
```diff
@@ -0,0 +1,11 @@
+testing
+testClickOnResultItem
+	| clickEvent hand position newBrowser oldBrowsers |
+	self runQuery: 'ALGCore'.
+	hand:= self algInstance world activeHand.
+	position:= self algInstance itemList submorphs second bounds center.
+	clickEvent:=MouseButtonEvent new setType: #mouseDown position: position which: 4 buttons: 4 hand: hand stamp: nil.
```
BenBals: Formatting: Line seems to long, missing spaces.

packages/AlgernonTests.package/ALGTestUI.class/instance/testMouseItemSelection.st
```diff
@@ -0,0 +1,9 @@
+testing
+testMouseItemSelection
+	| hand position moveEvent |
+	self runQuery: 'ALGCore'.
+	hand:= self algInstance world activeHand.
+	position:= self algInstance itemList submorphs third bounds center.
+	moveEvent:=MouseButtonEvent new setType: #mouseMove position: position which: 0 buttons: 0 hand: hand stamp: nil.
```
BenBals: Same formatting and code here. Maybe we could have a helper/fixture for creating these events, that sets defaults and you just overwrite what you need?

BenBals (changes requested):
I did not have a chance to run this code (have to be on Windows because of BIS), but the implementation seems really odd and suprisingly manual. Are you sure we have to do the mapping of events to items that way?

packages/Algernon.package/ALGCore.class/instance/mouseDown..st
sinnaj-r: Afaik we need the ALGCore to have Mouse-Focus to support the moving of Algernon

BenBals: Maybe we can support moving algernon only by clicking on the breadcrumbs and not the results?\

I think this would be a better solution.

packages/Algernon.package/ALGCore.class/instance/handlesMouseMove..st
sinnaj-r: I did not find a hover method/event.

BenBals: You can use `Morph>>mouseEnter:` and `Morph>>mouseLeave:`

BenBals:
Is it just me or does clicking feel slower than pressing enter on a result?

BenBals:
Could you hide the results for ? and help when the user has already typed one of these into the search bar? They should have figured it out by then.

BenBals:
The functionality seems as described though.

Issue #136, opened 2020-07-30
Fix Linting Issues


BenBals:
This is a large meta-issue that has many problems. The problems should be divided up.

- [x] Uses do: instead of collect: or select: (Weight: 4) @BenBals #138 
- [x] *Dot After Return Consistency* (Weight: 3) @robertkirsten #150 
      _(Please also check what the style guide says about last-line-dots without a return statement)_
- [x] ifTrue:/ifFalse returns instead of and:/or: (Weight: 1) @phlprcks -> was already refactored a while ago
- [x] intensive Coupling (Weight: 7) @sinnaj-r @BenBals #138
- [x] Law of Demeter (Weight: 9) @sinnaj-r @BenBals  #138
- [x] Long Methods (Weight: 4) @sinnaj-r #139
- [x] Sends questionable message (Weight: 1) @kej-jay 
- [ ] Sends add:/remove: to external collection (Weight: 1) @BenBals #146 
- [x] excessive number of arguments (Weight: 1) @phlprcks -> see comment below
- [x] Methods implemented but not sent (Weight: 1) @phlprcks #160 
- [x] Inconsistent method classification (Weight: 3, easy) @phlprcks -> see comment below
- [x] Temporary Variables not read and written (Weight: 3) @robertkirsten #153 
- [x] Instance Variable not read and written (Weight: 2) @robertkirsten #153 
- [x] Refers to class name instead of "self class" (Weight: 0.1) @kej-jay 
- [x] Unnecessary assignment to a temporary variable (Weight: 1) @robertkirsten #153
- [x] Unreferenced Variable (Weight: 3) @robertkirsten  #158 
	- The weight is so high, because the linter does not tell you which one (as far as I can tell)

| Team Member | Points |
|--|--|
| @sinnaj-r | 20 |
| @robertkirsten | 14 |
| @phlprcks | 6 |
| @kej-jay | 1.1 |
| @BenBals | 21 |

The weights are my subjective, imperfect measure of how time consuming and/or difficult a refactoring is as well as how much risk for escalation of time is involved. I suggest that each pair take on roughly equal amounts of weight. I welcome changes and more linting problems to be added.

This does not invalidate the "boy scout rule" that when you see one of these problems in the code you *must fix it immediatly*. That holds especially for simple refactorings like "Uses do: instead of collect: or select:".

sinnaj-r:
[...]

PR #139, opened 2020-07-31
Refactor Long Lines & Methods

sinnaj-r:
Most of the Long Methods & Lines - which _should_ be refactored - are refactored now, for the remaining ones I expect that other issues/refactorings will make changes anyway. A "shortening" of the methods therefore only makes sense afterwards.


packages/Algernon.package/ALGBackground.class/instance/initialize.st
```diff
@@ -1,11 +1,9 @@
...
-		center: World center;
-		borderWidth: 0.
+	self color: self class backgroundColor;
+		 extent: World extent;
+		 center: World center;
+		 borderWidth: 0.
 	self currentWorld addMorphFront: self.
-	self hide.
\ No newline at end of file
+	self hide
```
BenBals, line 9: Shouldn't there be a dot here?

packages/Algernon.package/ALGRepositoryType.class/class/algertemListWith..st
```diff
@@ -1,13 +1,5 @@
...
-		algertems add: (ALGItem withAlgertype: (self
-			withType: (each description)
-			withPackage: aWorkingCopyLabel))].
-	^ algertems
-	
-
+algertemListWith: aWorkingCopyLabel 
+	^ (self getReposFor: aWorkingCopyLabel)
+		collect: [:each | ALGItem
+				withAlgertype: (self withType: each description withPackage: aWorkingCopyLabel)]
```
BenBals, line 5: This might lead to a merge conflict with my `do:` refactorings, but we will care when it comes to that.

BenBals (approved):
Looks really good to my. Nice work. :+1:

packages/Algernon.package/ALGBackground.class/instance/initialize.st
sinnaj-r: Isn't our policy to omit the dot at the last line ?

packages/Algernon.package/ALGRepositoryType.class/class/algertemListWith..st
sinnaj-r: Alright 👍

packages/Algernon.package/ALGBackground.class/instance/initialize.st
BenBals: I think only after returns. But I'm not sure. Just do what you think is right.

sinnaj-r: I will keep it this way and add a note to #136 . 🙂

PR #150, opened 2020-08-03
Fix #136 Dot after return consistency

### LINKED ISSUE #136: Fix Linting Issues

robertkirsten:


packages/Algernon.package/ALGNullItem.class/instance/name.st
```diff
@@ -1,3 +1,3 @@
 accessing
 name
-^ 'Null Item'
\ No newline at end of file
+	^ 'Null Item'
```
BenBals, line 3: Are you sure there should be a tab here?

Maybe it's just github displaying this weirdly.

packages/Algernon.package/ALGResultList.class/instance/hasItems.st
```diff
@@ -1,3 +1,3 @@
 initialization
 hasItems
-^ self items notEmpty
\ No newline at end of file
+	^ self items notEmpty
```
BenBals, line 3: same

packages/Algernon.package/ALGResultList.class/instance/items.st
```diff
@@ -1,3 +1,3 @@
 event handling
 items
-^self submorphs
\ No newline at end of file
+	^ self submorphs
```
BenBals, line 3: same

BenBals (approved):
Looks good to me. :+1: Nice work.

packages/Algernon.package/ALGNullItem.class/instance/name.st
robertkirsten: All other return statements are also indented with a tab, therefore I don't see why there shouldn't be one.

BenBals: Okay, thank you, then it's just github.

BenBals:
As this is just a bunch of dots, I think we can merge without further review.

PR #153, opened 2020-08-03
Fix #136 Removed temporary/instance variables that are not read and written, unnecessary assignments

### LINKED ISSUE #136: Fix Linting Issues

robertkirsten:


packages/AlgernonTests.package/ALGTestCache.class/instance/testCacheKeyGenerationLevel1.st
```diff
@@ -1,7 +1,5 @@
 testing
 testCacheKeyGenerationLevel1
-
-	| parent |
-	parent := ALGItem new.
+	ALGItem new.
```
robertkirsten: I am not sure whether this is necessary for the test.

packages/AlgernonTests.package/ALGTestAction.class/instance/setUp.st
```diff
@@ -1,4 +1,3 @@
 running
 setUp
-	actions := ALGAction createListOfAlgeractionsGiven: (ALGClassType with: ALGCore).
-	self core: ALGCore new.
```
robertkirsten, line 4: I'm not 100% sure with this one, but from my testing removing this line doesn't alter the behavior. I modified a method in Algernon in such a way that the assertion fails, and it also did without this line. Plus removing this line causes the screen not to go white anymore when the tests in ALGTestAction are run.

packages/AlgernonTests.package/ALGTestKeyboard.class/instance/tearDown.st
```diff
@@ -1,5 +1,3 @@
 running
 tearDown
-
-	super tearDown.
-	toolWindow ifNotNil: [ toolWindow delete ]
```
robertkirsten, line 5: Here too, the line seems to be unnecessary.

packages/AlgernonTests.package/ALGTestCache.class/instance/testCacheKeyGenerationLevel1.st
BenBals: It should not be, a similar thing got removed here: https://github.com/hpi-swa-teaching/Algernon-Launcher/pull/155/files#diff-b1bbfaab62f80e23ef1bc09277ff9fd5

packages/AlgernonTests.package/ALGTestKeyboard.class/instance/tearDown.st
BenBals: Yes, I think this is unnecessary. Please delete the method as that will be the same as calling `super tearDown`.

packages/AlgernonTests.package/ALGTestAction.class/instance/setUp.st
BenBals: Seems good!

packages/AlgernonTests.package/ALGTestKeyboard.class/instance/tearDown.st
BenBals: Jep, please remove the method all together and remove the instance variable.

robertkirsten:
Do you have any idea why the CI run fails @BenBals?

BenBals:
No, maybe @sinnaj-r has an idea.

robertkirsten:
Works now.

.github/workflows/smalltalk-ci.yml
```diff
@@ -23,4 +23,4 @@ jobs:
       shell: bash
       timeout-minutes: 15
       env:
-        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        COVERALLS_REPO_TOKEN: ${{ secrets.COVERALLS_REPO_TOKEN }}
```
sinnaj-r: I have the feeling, that this change is not intended. (Also regarding to the not working CI)

robertkirsten: This change is from a commit from @fniephaus which was added when I merged from development

robertkirsten: Or is it? Wait I need to check

robertkirsten:
So we are ready to merge?

README.md
```diff
@@ -1,6 +1,4 @@
-# Algernon Launcher
```
sinnaj-r: This Looks like you are reverting changes here 🤔

sinnaj-r (approved):
Theres one strange readme change, but apart from that this can be merged 👍 
(After resolving the merge-conflicts)

BenBals (approved):
Lgtm

PR #155, opened 2020-08-04
Implement #80 v2

phlprcks:
finalizes #80 (on a new branch bc the git browser seemingly does not work well with pushing a rebased branch...)


packages/Algernon.package/ALGItem.class/instance/nameLessThan.given..st
```diff
@@ -4,6 +4,4 @@ nameLessThan: anItem given: aLowercaseString
 	| me other |
 	me := self lowerCaseName beginsWith: aLowercaseString.
 	other := anItem lowerCaseName beginsWith: aLowercaseString.
-	(me and: [other not]) ifTrue: [^ true].
-	(other and: [me not]) ifTrue: [^ false].
-	^ #equals
\ No newline at end of file
+	^(me and: [other not]).
```
BenBals: This code is almost the same as `initalsLessThan:given:` please factor out the duplication.

packages/Algernon.package/ALGItemFilter.class/instance/accepts..st
```diff
@@ -2,5 +2,5 @@ filtering
 accepts: aStringCollection
 
 	(aStringCollection size <= 1)
-		ifTrue: [ ^ false ].
-	^ self class keywords includes: aStringCollection first
\ No newline at end of file
+		ifFalse: [ ^ self class keywords includes: aStringCollection first ].
+	^ false
```
BenBals: This should be `(aString collection size > 1) and: [ self class keywords includes: aStringCollection first ]`

EDIT: Maybe include a line break.

BenBals (changes requested):
Thank you very much! This looks like a big and noticeable improvement to the code quality! :1st_place_medal: 

I just have two little nitpicks, nothing big.

phlprcks:
@BenBals or @sinnaj-r any ideas why squeak 5.3 on macos-latest has a timeout for testDailyCacheCleanup. That is the only fail - everything else runs smoothly

BenBals:
This happens all the time and on other branches too. Maybe @sinnaj-r has an idea. It would be cool to see this fixed.

packages/Algernon.package/ALGItem.class/instance/initialsLessThan.given..st
```diff
@@ -1,7 +1,8 @@
 comparison
 initialsLessThan: anItem given: aLowercaseString
 
-	| me other |
+	"| me other |
 	me := self initials beginsWith: aLowercaseString.
 	other := anItem initials beginsWith: aLowercaseString.
-	^(me and: [other not])
+	^(me and: [other not])"
```
BenBals: Is there a reason why you left this in as a comment? If not please delete.

packages/Algernon.package/ALGItem.class/instance/nameLessThan.given..st
```diff
@@ -1,7 +1,8 @@
 comparison
 nameLessThan: anItem given: aLowercaseString
 
-	| me other |
+	"| me other |
```
BenBals: Dito.

BenBals (changes requested):

sinnaj-r:
> @BenBals or @sinnaj-r any ideas why squeak 5.3 on macos-latest has a timeout for testDailyCacheCleanup. That is the only fail - everything else runs smoothly

I think I've identified the problem: We only increased the timeout for the ALGTest class (+subclasses), but not for the ALGTestInstallation class.
This is not a problem for the other non-ALGTest classes, as they don't execute much logic; however, the testDailyCacheCleanup test does a complete Algernon-Cleanup, which usually takes a long time. 

So simply add following method to the ALGTestInstallation class
```
defaultTimeout

	^ 60 "seconds" 
```

BenBals:
I did that, see #156

sinnaj-r (approved):
Apart from the the two remaing comments - see Bens remark - this looks good 👍

phlprcks:
@BenBals thanks for making me aware of the outcommented parts - that was not on purpose and I oversaw that. Please feel free to merge and close the issue once you can approve :)

BenBals (approved):
Looks good now.

PR #156, opened 2020-08-04
Increase Timeout for Installation Tests

BenBals:


phlprcks:
As it was needed for testing and included only minor changes that looked good, I just merged it.

PR #157, opened 2020-08-04
Fix #130

### LINKED ISSUE #130: Fix the =-Space Expression Bug
At the Moment the ALGExpressionType>>getCommand Method will cut the first two characters from the input and run the rest. That means that only "= [EXPRESSION]" will work not "=[EXPRESSION]".
This should be an quite easy fix we should do before submission.

Remember:
- [x] Write a regression test

kej-jay:


BenBals (approved):
Looks good to me.

packages/Algernon.package/ALGExpressionType.class/instance/getCommand.st
```diff
@@ -1,6 +1,9 @@
 actions
 getCommand
-	| textFieldContents |
+	| textFieldContents startIndex |
 	textFieldContents := self typeObject topmostQuery.
-
-	^ textFieldContents copyFrom: 3 to: textFieldContents size"first 2 characters are do or '= '"
\ No newline at end of file
+	(textFieldContents at: 1) == $=
```
sinnaj-r: Just some nitpicking:
`textFieldContents first = $=`
 would be a bit more squeak-like

sinnaj-r (approved):
Looks good 👍 
Just got one small comment, after that you should be fine to merge

packages/Algernon.package/ALGExpressionType.class/instance/getCommand.st
kej-jay: did that :+1:

sinnaj-r: Thank you 👍

kej-jay:
Will merge this now!

PR #158, opened 2020-08-04
Fix #136 Removed unreferenced variables

### LINKED ISSUE #136: Fix Linting Issues

robertkirsten:


robertkirsten:
Ready to merge?

BenBals (approved):
Good job. :rocket:

PR #160, opened 2020-08-05
refactor linting parts of #136

phlprcks:
see comments in #136 for details

PR #162, opened 2020-08-05
removed not needed methods

phlprcks:


phlprcks:
@sinnaj-r as there are difficulties with the readme and yaml files even after rebasing in squeak git browser, I felt saver to implement it in this new branch instead of  #160 .

sinnaj-r (approved):

PR #163, opened 2020-08-05
Fix #134 - Custom assertions

### LINKED ISSUE #134: Customs asserts and helpers in test code
Much of the test code is duplicated. We can use custom asserts and helpers to clean that up.

robertkirsten:


packages/AlgernonTests.package/ALGTestCache.class/instance/testCacheKeyGeneration.st
```diff
@@ -1,6 +1,5 @@
 testing
 testCacheKeyGeneration
```
phlprcks, line 2: isn't there a way to avoid the code duplication with `testCacheKeyGenerationLevel0` and potentially even more methods (e.g. `testCacheKeyGenerationLevel1`)?

packages/AlgernonTests.package/ALGTestTypes.class/instance/testExpressionDoNoTranscript.st
```diff
@@ -1,6 +1,5 @@
 testing expressions
 testExpressionDoNoTranscript
```
phlprcks, line 2: generally I think the newline is not considered to be bad practice - but as you changed it we can just leave it as it is :)

packages/AlgernonTests.package/ALGTestTypes.class/instance/testPackageAction.st
```diff
@@ -5,5 +5,5 @@ testPackageAction
 	self moveCursorTo: 'Package'.
```
phlprcks: wasn't there a helper for `runQuery .. withCursorAt` or sth like that 🤔

phlprcks (changes requested):
looks like the tests have much improved! good job 👍 
I added minor comments to some files - apart from that it looks fine for me

packages/AlgernonTests.package/ALGTestTypes.class/instance/testExpressionDoNoTranscript.st
robertkirsten: Yes.

packages/AlgernonTests.package/ALGTestTypes.class/instance/testPackageAction.st
robertkirsten: I don't see it? What would it be called?

packages/AlgernonTests.package/ALGTestCache.class/instance/testCacheKeyGeneration.st
robertkirsten: I think it's not really worth it as still many parameters would need to be passed and the there are not enough duplications.

packages/AlgernonTests.package/ALGTestTypes.class/instance/testPackageAction.st
phlprcks: `runQuery: query withCursorAt: aCategory` could be used to refactor line 4-5 if I am not mistaking.

robertkirsten: I didn't get your question and didn't look at the code, I thought you meant `runQuery:withCursorAt:` is already implemented. Yes it can be used here

packages/AlgernonTests.package/ALGTestAction.class/instance/assertTypeCreateListOfAlgeractions..st
```diff
@@ -0,0 +1,3 @@
+helpers
+assertTypeCreateListOfAlgeractions: aType
```
sinnaj-r: I think this name could be more descriptive. Something like `assertAlgeractionsNotEmptyForType:`

packages/AlgernonTests.package/ALGTestTypes.class/instance/denyTypeIsEmpty..st
```diff
@@ -0,0 +1,3 @@
+helpers
+denyTypeIsEmpty: aType
```
sinnaj-r: I would rename this as well. Something like `denyEmptyAlgertemListForType:`

sinnaj-r (approved):
I really like the new helpers and general cleanup you did!👍 

There were just two helpers I would have named differently

PR #176, opened 2021-05-10
fix: use correct specs for CI

FranzSw:
- updates baseline to fix failing CI


TheGrayStone (approved):
Looks like the changes work - can be merged.

PR #180, opened 2021-05-14
Ci: add coverage

FranzSw:
Because of the BaselineTests, coverage has been disabled.
This PR removes the deprecated attribute `smalltalk-version` in the workflow and adds a new workflow that only executes AlgernonTests and sends the coverage to coveralls.

(I hope that we can get the coveralls bot to comment the coverage on PRs, but I think an SWT-admin has to change that)

### DIFF
--- a/.coverage.ston
+++ b/.coverage.ston
@@ -13,7 +13,7 @@ SmalltalkCISpec {
   ],
   #testing : {
     #coverage : {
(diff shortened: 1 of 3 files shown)


TheGrayStone (approved):
It looks about right and I applaud what you did to get it to work 👏

PR #181, opened 2021-05-19
Feat/menu commands updating

### LINKED ISSUE #175: Support more (squeak) menu commands
As a user I want to find as many commands as possible in Algernon in order to not click through menus

## Acceptance Criteria
- [x] Make all top-level menu bar items accessible in Algernon
- [x] don't forget to test

niklasmohrin:
Closes #175


.github/workflows/smalltalk-ci.yml
```diff
@@ -5,22 +5,20 @@ jobs:
...
+    - uses: actions/checkout@v2
+    - uses: hpi-swa/setup-smalltalkCI@v1
+      id: smalltalkci
+      with:
+        smalltalk-version: ${{ matrix.smalltalk }}
+    - run: smalltalkci -s ${{ steps.smalltalkci.outputs.smalltalk-version }}
+      shell: bash
+      timeout-minutes: 15
+      env:
+        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
FranzSw: This is an old version (`smalltalk-image` is deprecated), I guess the changes were accidental or Squeak does weird stuff again?

FranzSw (changes requested):
Not sure why, but `.coverage.ston`, `coverage-reporter.yml` are deleted and the issue below.

FranzSw (approved):

bewee (approved):

packages/Algernon.package/ALGWorldMenuType.class/class/filterMenuItems..st
```diff
@@ -0,0 +1,4 @@
+helpers
+filterMenuItems: aMorph
```
niklasmohrin: hm, there isn't really any filtering of menu items going on, is there?

what about `scrapeMenuItems`? I fell like "filter" / "select" reads as if the `aMorph` was a collection, though it isn't (and the name `each` doesn't make that clear either)

packages/Algernon.package/ALGWorldMenuType.class/class/menuEntries.st
```diff
@@ -0,0 +1,17 @@
+helpers
```
niklasmohrin: Why "helpers" exactly? It seems to me that this is _strongly_ tied to the generation of the algtermlist and having the method in it's own category doesn't do code navigation any good either

packages/Algernon.package/ALGStaticWorldMenuType.class/class/filterMenuItems..st
```diff
@@ -0,0 +1,8 @@
+helpers
+filterMenuItems: aMorph
+	| entries |
+	entries := OrderedCollection new.
+	(aMorph isKindOf: DockingBarUpdatingMenuMorph) 
+		ifFalse: [ entries addAll: aMorph submorphs ].
+						
+	^ entries
```
niklasmohrin:
```suggestion
helpers
filterMenuItems: aMorph
	^ (aMorph isKindOf: DockingBarUpdatingMenuMorph) 
		ifTrue: [ OrderedCollection new ]
		ifFalse: [ OrderedCollection newFrom: aMorph submorphs ].
```

niklasmohrin (commented):
Sorry for the late review :sweat_smile:

packages/Algernon.package/ALGWorldMenuType.class/class/filterMenuItems..st
Olliwehr: Yeah, you're right. I would suggest sth like `extractMenuEntriesFromMenuMorph` (or `extractMenuEntriesFrom: aMenuMorph` for short) as we do exactly that, no?

packages/Algernon.package/ALGWorldMenuType.class/class/menuEntries.st
Olliwehr: This then also applies to the other methods in this class (`menus`, `filterMenuItems`, `dockingBar`), doesn't it?
So what about putting all those in the category `algertem lists`?

packages/Algernon.package/ALGStaticWorldMenuType.class/class/filterMenuItems..st
Olliwehr: Quite a nice suggestion. The current implementation seems to be similar in structure to the same method in `ALGDynamicWorldMenuType`. Should we also adjust the implementation there or is that not so relevant?

packages/Algernon.package/ALGWorldMenuType.class/class/menuEntries.st
bewee: I had a look at the other types. Some put their algertem list related methods into the category `algertem list`, others put them into other categories like `helper` or `instance creation`. So I think putting them all into `algertem list` and removing the `helpers` category is fine.

packages/Algernon.package/ALGStaticWorldMenuType.class/class/filterMenuItems..st
bewee: I think so

bewee:
Why does this **always** happen when I push a commit?😬😓😤

packages/Algernon.package/ALGDynamicWorldMenuType.class/class/extractMenuEntriesFrom..st
```diff
@@ -0,0 +1,9 @@
+algertem lists
+extractMenuEntriesFrom: aMorph
+	^ (aMorph isKindOf: DockingBarUpdatingMenuMorph) 
+		ifTrue: [ | entries |
+				aMorph updateMenu. 
+				entries := aMorph submorphs.
```
niklasmohrin: From `Morph>>submorphs`:

> This method returns my actual submorphs collection. Modifying the collection directly could be dangerous; make a copy if you need to alter it.

I guess it should be `OrderedCollection newFrom: aMorph submorphs` to be safe

niklasmohrin (commented):
Sorry for nitpicking, I am gonna go ahead and commit this one myself :smile:

LinqLover:
> Why does this **always** happen when I push a commit?😬😓😤

This is unfortunately a known bug in Squot: https://github.com/hpi-swa/Squot/issues/199 Usually it should help to commit everything before pulling (technically, this allows the pull to act as a fast-forward merge).

bewee:
So can we now merge this?

PR #179, opened 2021-05-12
Feat/support scrolling results

bewee:

Issue #169, opened 2021-04-30
Support scrolling result list


FranzSw:
As a user I want to be able to scroll the result list in order to see more results

## Acceptance criteria
- [x] Add a configuration item for result items limit (default: 20 items)
- [x] If there are more results than can be shown at once, enable scrolling
- [x] The configured amount of search results should be accessible in the list

PR #184, opened 2021-05-25
Follow-Up-Work for Feat/support scrolling results

Olliwehr:
- Little test adjusting as our coverage has decreased after merging (which meant a red build on development)
- references #169 and #179


TheGrayStone (approved):
Seems to get the job done without much fuzz.

bewee (approved):
👍

PR #195, opened 2021-05-28
Feat/window actions

### LINKED ISSUE #174: Support actions for open windows
As a power user of Algernon, I want to be able to search for open windows and execute actions on them in order to speed up my workflow

## Acceptance Criteria
- [x] Show open windows in search results
- [x] Show actions for windows:
- [x] bring to front
- [x] expand (+)
- [x] contract (+)
- [x] collapse (-) 
- [x] close (x)
- [x] close all like this
- [x] close all but this

bewee:
Closes #174


packages/Algernon.package/ALGActionVerbFilter.class/class/keywords.st
```diff
@@ -2,4 +2,6 @@ filtering
 keywords
 	^ Set newFrom: {'run'. 'browse'. 'fileout'. 
 		'open'. 'saverepo'. 'do'. 
-		'inspect'. 'explore'. '=' }
\ No newline at end of file
+		'inspect'. 'explore'. '='.
+		'front'. 'expand'. 'collapse'.
```
Olliwehr, line 6: I think it would be nice if we add some helptexts for those new keywords in the `helpTexts` method that is also located in the ActionVerbFilter class. Adding those explainations to the [Keywords Wiki Page](https://github.com/hpi-swa-teaching/Algernon-Launcher/wiki/Keywords) would be the next step then :)

packages/Algernon.package/ALGWindowType.class/class/algertemList.st
```diff
@@ -0,0 +1,5 @@
+as yet unclassified
```
Olliwehr: This method and also `visibleWindows` seem to be unclassified at this point. I thought we had a linting test for this but maybe the test for no unclassified methods just covers instance methods? 🤔

packages/Algernon.package/ALGWorldMenuType.class/class/menuEntries.st
```diff
@@ -5,11 +5,11 @@ menuEntries
 	entries := OrderedCollection new.	
 
 	((self menus
-		collect: [ :each | each subMenu ])
-		reject: [ :each | each isNil ])
+		collect: [ :each | {each subMenu. each contents} ])
+		reject: [ :each | each first isNil ])
 		collect: [ :each |
```
Olliwehr: I know that the original implementation came from another issue, but do we really collect sth here? As we just add entries to an outer local variable, a do instead of collect would have the same effect right?

packages/AlgernonTests.package/ALGTestTypes.class/instance/testWindowCollapse.st
```diff
@@ -0,0 +1,10 @@
+testing window type
+testWindowCollapse
+
+	|browser|
+	
+	browser := self openBrowser.
+	self testOpenWindowEntry: 'System Browser' withAction: 'collapse'.
+	Transcript show: browser isCollapsed.
```
Olliwehr: Is this Transcript printing still intended or was it just for debugging? ^^

Olliwehr (changes requested):
So, i have commented on some minor things.
But definitively big 👍 for those testcases.

packages/AlgernonTests.package/ALGTestTypes.class/instance/testWindowCollapse.st
bewee: 100% intended :D

packages/Algernon.package/ALGWorldMenuType.class/class/menuEntries.st
bewee: You're right, absolutely nothing is collected here, and luckily we never touched this collected nothing🤣

I just feel like `extractMenuEntriesFrom:` is still not quite the right name, since the extracted menu entries may contain line spacers or some other bloat, which won't be filtered out until the very end of the `menuEntries` function in `entries select: [:each | each isKindOf: MenuItemMorph]`

packages/Algernon.package/ALGWindowType.class/class/algertemList.st
bewee: Hm, very weird. Maybe we should have another look at these linting tests?

packages/Algernon.package/ALGActionVerbFilter.class/class/keywords.st
bewee: Ah, you're right, we totally forgot about these🙈
I can add these to the wiki page as soon as this PR is merged if that's okay?

packages/Algernon.package/ALGWindowType.class/class/algertemList.st
bewee: Ok, I feel like this is a problem with SWALint. And I either couldn't find any other lint which checks for class method categorization :/ Maybe you want to have another look at that?

Olliwehr: Okay, that is unfortunate. Yep, as far as i can see there is no equivalent test for class methods.

packages/Algernon.package/ALGWorldMenuType.class/class/menuEntries.st
Olliwehr: Ahh, i didn't note that the line spacers and other bloat won't be filtered out until later. Then Niklas' "scrape" suggestion from the other PR should fit better. Sth like `scrapeMenuItemsFrom` as "MenuItems" includes those bloat items more then "MenuEntries" and scraping underlines that there is bloat included?

Olliwehr (approved):

packages/Algernon.package/ALGWorldMenuType.class/class/menuEntries.st
bewee: @niklasmohrin Thoughts?

TheGrayStone (approved):
Looks good now. Please don't forget to add the new keywords to the wiki once merged.

packages/Algernon.package/ALGWorldMenuType.class/class/menuEntries.st
niklasmohrin: Yes I have foreseen this miles away and am very pleased that you have finally given in /s

Yeah, that all sounds right to me 👍

PR #198, opened 2021-05-31
Scrolling input field

### LINKED ISSUE #94: The Text-Input should "scroll" with the cursor
As a user I want to see the entered text also for a long string (e.g.  e.g. a Smalltalk-Expression.) in the Input Field in order to navigate and check my typed letters in the input field.

Additional Info:
Currently the textfield doesn't show the text around the current cursor position, but only the first words (which are then truncated).

# Acceptance Criteria
| Team | Customer | Title |
|------|----------|-------|
| :x: | :x: | Text around the curser is shown in the input field |
| :x: | :x: | the text "moves/scrolls" when text that exceeds the field size is entered |
| :x: | :x: | the text "moves/scrolls" according to the cursor position (e.g. on navigating with the cursor|

niklasmohrin:
Closes #94


packages/Algernon.package/ALGInputWrapper.class/instance/handleKeystroke..st
```diff
@@ -0,0 +1,3 @@
+as yet unclassified
+handleKeystroke: anEvent 
+	" no-op "
```
niklasmohrin: We should probably whitelist methods we want to forward in `doesNotUnderstand` instead of blacklisting like this; it's something we forgot yesterday

TheGrayStone: Should be done now in the newest commit. Please check.

TheGrayStone:
Ok. Now selecting all of the last query on open doesn't work anymore. Someone should check up on that.

packages/Algernon.package/ALGInputWrapper.class/instance/handleKeystroke..st
```diff
@@ -0,0 +1,4 @@
+forwarding
+handleKeystroke: anEvent 
+
+	" no-op "
```
niklasmohrin: I don't really understand what this is doing and future devs won't either, so maybe a comment should be added here?

TheGrayStone (approved):

packages/Algernon.package/ALGInputField.class/instance/normalKeyStroke..st
```diff
@@ -4,6 +4,8 @@ normalKeyStroke: event
 	"the controller notifies the text field of symbols entered,
 	special keys are handled by the controller"
 	self editor hasSelection ifTrue: [^super keyStroke: event].
-	
 	super keyStroke: event.
+	World doOneCycle.
+	self changed: #normalKeystroke.
+
```
bewee:
```suggestion
	self changed: #normalKeystroke
```

bewee (approved):
Only stumbled across this one trailing dot, the rest looks good to me.
Let's merge?

PR #202, opened 2021-06-07
Feat/reload classes

### LINKED ISSUE #190: Dynamically reload classes when creating new ones
As a programmer using Algernon I want every new class to automatically show up in the results without having to reset Algernon, in order to make my workflow more convenient.

## Acceptance criteria
- [x] Add every new class to the item list not more than 1 minute after creation
- [x] Remove every deleted class from the item list not more than 1 minute after deletion

Olliwehr:
closes #190


knollengewaechs:
The only test that is not running yet is testEnableAfterInstall in BOALGTest, because it installs Algernon from the development branch and doesn't find the new message yet. This will most likely be fixed with the merge of this PR into development

packages/Algernon.package/ALGItemProvider.class/instance/handleSystemChangeEvent..st
```diff
@@ -0,0 +1,4 @@
+updating
+handleSystemChangeEvent: anEvent
+	self class initializeItems.
+	self cache clear
```
bewee, line 4: I'm not really sure whether it's really such a great idea to handle it like this. On the other hand, everything still runs smoothly (again kinda proving the uselessness of this weird cache??). 
Anyway, I feel like the appropriate solution would be to make the ClassType dynamic. But then again ... This would ultimately disable the cache for all classes, whereas I guess we can expect the manual cache clear because of the creation/removal of a class not to happen too often. It's a difficult consideration between the harm done by not caching classes vs trashing the whole cache (kinda) regularly😕
In the end, I feel like it pretty much does not matter because we happen to have very few benefits from the cache anyway...
What do you think?🤔

packages/AlgernonTests.package/ALGTestClassReload.class/instance/setUp.st
```diff
@@ -0,0 +1,6 @@
+running
+setUp
+
+	super setUp.
+	className := #ALGDummyClass.
+	self deleteClass.
```
bewee:
```suggestion
	self deleteClass
```

bewee (commented):
First of all: The feature works pretty good, and I really like the tests - Great work here👍👍

Although that's not explicitly stated in the issue, I think that - since the intent was not having to manually reset Algernon during programming - reloading packages and methods is part of this issue? I also feel like this should be pretty easy to implement now, isn't it? (even the tests, I think a lot from `testCreateAndDeleteClass` could be reused)? Or do you think that's rather another issue?

Tbh, I would prefer to have one more person reviewing this because ot the cache issue mentioned below. I will quickly assign @FranzSw and hope that's okay😅

packages/Algernon.package/ALGItemProvider.class/instance/handleSystemChangeEvent..st
Olliwehr: Yeah, it feels more like removing a symptom by clearing the whole cache than really dealing with the root cause (locating which parts of the cache need to be exactly cleared when new classes appear). But as far as my understanding of the fuzzy matching goes, we lookup the cache for every new character that is typed in the search bar in order to achieve sth like substring-matching. And at least for me that makes it kinda unpredictable which parts of the cache really need to be cleared.

As you noted adding/removing classes won't happen that much, so that was our consideration for just clearing the whole cache on those occasions.

Olliwehr:
@bewee 
> Although that's not explicitly stated in the issue, I think that - since the intent was not having to manually reset Algernon during programming - reloading packages and methods is part of this issue? I also feel like this should be pretty easy to implement now, isn't it? (even the tests, I think a lot from `testCreateAndDeleteClass` could be reused)? Or do you think that's rather another issue?

We favoured packing that into new user stories as only reloading classes was requested (tbh the issue originally didn't even took reloading when classes get deleted into consideration). As you noted a lot of our code can be reused and e.g. when taking categories and methods into consideration, those kinds just need to be added into the `itemKinds` in `subscribeToSystemChangeNotifier` and there you go.

What i am trying to say: those new user stories won't produce much effort (but we could pose like that for the customer? 😄)

packages/AlgernonTests.package/ALGTestClassReload.class/instance/setUp.st
bewee: @Olliwehr Fyi, you could just have clicked the "commit suggestion" button here, then you will not have to have created a new commit (Futur II in English is hard😅)

Olliwehr: @bewee i did exactly that and just edited the commit message. ^^ But as far as i noticed that will always produce a new commit, no?

bewee: @Olliwehr Ahh ok, I'm sorry then. You confused me by knowing that one can change the commit message there😅😂
Always nice to learn new stuff :D

bewee (approved):

bewee:
@FranzSw Olli and I talked and decided that it's fine for now. But feel free to have a look at it if you still want to😅

PR #203, opened 2021-06-07
Feat/numbers

### LINKED ISSUE #189: Add key combos to instantly access the first few results of a query
As a power user of Algernon I want to be able to access the first few results of a query without having to focus on them via scrolling in order to speed up my workflow.

## Acceptance criteria
- [x] Add a customizable keyboard shortcut (default cmd)
- [x] Show numbers next to first 9 results (not affected by scrolling)
- [x] Trigger result number `x` when entering `shortcut` + Number key `x`

TheGrayStone:
Closes #189


packages/Algernon.package/ALGCore.class/instance/ifShouldHandleAsQuickAccess.do..st
```diff
@@ -0,0 +1,6 @@
+events-processing
+ifShouldHandleAsQuickAccess: anEvent do: aBlock
+    ^ ((anEvent modifierString sameAs: self class quickAccessModifier , ' ')
+            and: ['123456789' includes: anEvent keyCharacter])
+        ifTrue: [aBlock value]
+		ifFalse: [ nil ]
```
bewee:
```suggestion
    ((anEvent modifierString sameAs: self class quickAccessModifier , ' ')
            and: ['123456789' includes: anEvent keyCharacter])
        ifTrue: [aBlock value]
```
As the return value is never used, this should work as well.

packages/Algernon.package/ALGCore.class/class/setQuickAccessModifier..st
```diff
@@ -0,0 +1,6 @@
+preferences
+setQuickAccessModifier: aString
+    ({'cmd'. 'ctrl'} includes: aString asLowercase)
+        ifTrue: [self quickAccessModifier: aString.
+            ^ true].
+    ^ false
```
bewee:
```suggestion
    ^ ({'cmd'. 'ctrl'} includes: aString asLowercase)
        ifTrue: [self quickAccessModifier: aString. true].
        ifFalse: [false]
```
It feels like this is easier to understand. But 🤷‍♂️

bewee: Or could we maybe even remove the `ifFalse:` branch completely and it will still return false in this case because that's the default behaviour?

packages/Algernon.package/ALGItemProvider.class/instance/searchAllItemsFor..st
```diff
@@ -3,6 +3,7 @@ searchAllItemsFor: aString
 	| items dynamicItems |
 	items := self getCachedStaticItemsFor: aString.
 	items ifEmpty: [ items := self searchStaticItemsFor: aString ].
+"	items := self searchStaticItemsFor: aString."
```
bewee:
```suggestion
	items ifEmpty: [ items := self searchStaticItemsFor: aString ].
	
```

packages/Algernon.package/ALGResultList.class/instance/add.target.selector.argumentList..st
```diff
@@ -6,5 +6,7 @@ add: anItem target: target selector: aSymbol argumentList: argList
 				target: target
 				selector: aSymbol
 				argumentList: argList
-				in: self.
+				in: self
+				index: self currentItemIndex.
+	self currentItemIndex: self currentItemIndex + 1.
```
bewee: Tbh I don't really like the idea of having an instance variable that is only in use during initialization, but of course still there as soon as everything is finished. Especially since you've written public accessors for it.

I'd rather:
- Make it a private variable with no public accessors or
- Better: Rewrite it to use a local counter variable (in `updateWith:`), possibly removing the indirection via the `addAll:`

bewee (approved):
Since @TheGrayStone asked me to review this, I quickly self-assigned and marked it as ready.

The feature looks good and works fine, great work. I only had some minor remarks you might want to have a look at before merging.

packages/Algernon.package/ALGCore.class/class/setQuickAccessModifier..st
niklasmohrin: What about

```suggestion
setQuickAccessModifier: aString
    | valid |
    valid := ({'cmd' . 'ctrl'} includes: aString asLowercase).
    valid ifTrue: [ self quickAccessModifier: aString].
    ^ valid
```

packages/Algernon.package/ALGResultList.class/instance/add.target.selector.argumentList..st
TheGrayStone: Oof... I think this is pretty difficult to realize. If you think you can do it better, just push a commit.

packages/Algernon.package/ALGCore.class/instance/ifShouldHandleAsQuickAccess.do..st
TheGrayStone: What?

PR #209, opened 2021-06-10
Fix CI to use development branch again

bewee:
We had to do this because development and feat/reload-classes were incompatible, but we are denied to merge when tests won't pass.

Co-authored-by: Oliver Heß <oliver.hess@student.hpi.de>


Olliwehr (approved):
👍

PR #210, opened 2021-06-11
Release CI without automatic commit messages

FranzSw:
- CI Runner that creates a SAR and adds it to a release when pushing a tagged commit
- for now without automatic commit message collection


niklasmohrin (changes requested):
lets wait another day, works in the fork, not worth having this for a day

PR #211, opened 2021-06-12
Release Preparation: Bump Version

FranzSw:
- bumps the version by one

What should our release version be? 4.0.0?


bewee:
~~Why not just use sth like 3.1.0? Then we could keep the 4.0.0 for our final release at the end of the semester?~~

Edit: You're right, breaking changes. Let's do this!

bewee (approved):

PR #213, opened 2021-06-21
Execute single action events on tab

### LINKED ISSUE #35: Use Tab to execute result
As an Algernon power-user, I want to be able to use the Tab-key to execute the result when no further detail actions are available, so I don't have to mentally switch the key I have to press.

## Acceptance criteria:
- [x] If no further actions for an item are available and I hit the Tab-key, the item is executed.
- [x] Add a preference to toggle this behavior

bewee:
- closes #35


packages/Algernon.package/ALGCore.class/class/resetKeyBindings.st
```diff
@@ -0,0 +1,4 @@
+constants
+resetKeyBindings
```
Olliwehr: Is this method even used in the current code? Besides that, does it make sense to call it when we reset/cleanup Algernon?

packages/AlgernonTests.package/ALGTestItemList.class/instance/testTabExecutesSingleActionResult.st
```diff
@@ -0,0 +1,14 @@
...
+	
+	browserCount := (self findSystemWindowsByLabel: 'Preference Browser') size.
+
+	self runQuery: 'Preferences'.
+	
+	self algInstance keyStroke: (ALGFakeKeyboardEvent new keyString: '<tab>' shallowCopy;
+			 isKeyUp: true).
+			
+	browsers := self findSystemWindowsByLabel: 'Preference Browser'.
+	self assert: browsers size > browserCount
```
Olliwehr, line 15: Not sure if that will be a problem, but if you disable the introduced setting for the tab execution in the image, this and the next test fail (at least in my image). What about two branches of assertion (so if the setting is enabled then assert like you do now and when disabled then assert that no new browser gets opened)? 🤔

Olliwehr (changes requested):
Code looks good overall 👍 

I commented on some minor things, but let me also remind you of updating the README.md, as the meaning of using tab has changed.

packages/AlgernonTests.package/ALGTestItemList.class/instance/testTabExecutesSingleActionResult.st
FranzSw: I think we should move that to another test case. Because the CI always runs the tests isolated, the setting should not be a problem. However I guess it would be convenient to ensure the setting is set / set it and restore it to be able to run it locally successfully.

packages/Algernon.package/ALGCore.class/class/resetKeyBindings.st
bewee: Actually, they don't really ever change. We only added it because one key now had to map to another function. Latest commit removes it

packages/AlgernonTests.package/ALGTestItemList.class/instance/testTabExecutesSingleActionResult.st
bewee: Should be solved with the most recent commit

bewee:
Maybe we should move this pattern "Store preferences in setUp, change preferences in tests as needed, restore preferences in tearDown" to the ALGTest class and apply it to all preferences?
Edit: Obviously in another issue😅

Olliwehr (approved):
Moving that pattern to the general test class would be nice, but rather in another user story? At the moment i cant really imagine the amount of work that will produce.

But for now 👍 for the adjustments. Also ran the tests in case of the disabled settings in my image, it now works for me.

Do we still need the little adjustment in the README @bewee?

bewee:
@Olliwehr I completely forgot about the README😅 Anyway, I'm not sure whether we really have to edit it, since all the descriptions there are only about two words long, and by describing this feature, I wouldn't be able to keep it nearly as short 🙈 Also, you can deactivate it, and the feature is really subtle.

What do you think?

bewee:
> Moving that pattern to the general test class would be nice, but rather in another user story? At the moment i cant really imagine the amount of work that will produce.

That would be quite a bit of copy-and-paste. However, it's definitely not a user story, since there's no benefit for the user😅
Another idea I had for test refactoring (writing it here just so it's written down somewhere 🙈 ) is to rewrite all the "opens a window" tests to sth like
```
self assert: [ self query: 'Preferences'; keyTab ] opensWindow: 'Preferences'|
self deny: [ self query: 'Browser'; keyTab ] opensWindow: 'System Browser'|
```
Again providing these assertions in the `ALGTest` class

Olliwehr:
Okay, you convinced me regarding the README 😅 As the existing bevaviour of using tab is still navigating in most cases, yeah lets keep it like that. Then i think we can merge this! @bewee

PR #214, opened 2021-06-25
Refactor tests

FranzSw:
Restructures tests to separate unit & integration tests (what do you think?)


packages/Algernon.package/ALGClassType.class/class/exploreAllInstancesFor..st
```diff
@@ -0,0 +1,12 @@
+helpers
+exploreAllInstancesFor: aClass
+
+	| all allSize prefix |
+	all := aClass allInstances.
+	(allSize := all size) = 0 ifTrue:
```
niklasmohrin, line 6: `isZero`

packages/AlgernonTests.package/ALGIntegrationTest.class/instance/assertQuery.opensNewWindow..st
```diff
@@ -0,0 +1,14 @@
...
+
+ 	browsers := self findSystemWindowsByLabel: windowName.
+
+ 	self runQuery: query.
+	self activateCurrentItemAction.
+	
+	World doOneCycle.
+
+ 	newBrowsers := self findNewSystemWindowsByLabel: windowName old: browsers.
+ 	self assert: newBrowsers size = 1.
```
niklasmohrin, line 14: `assert:equals:`

FranzSw:
Because the semester comes to an end and we are not convinced that a different structure will help us with writing good tests, we decided to not tackle this now.

PR #221, opened 2021-07-04
Add world submenu entries and checkboxes

### LINKED ISSUE #187: Make submenus of world menu entries accessible
As an avid user of Algernon I want to be able to access submenus of world menu entries via Algernon, in order to fully replace using the world menu with Algernon.

## Acceptance criteria
- [ ] Add submenu entries for all existing world menu items in deeper hierarchy layers (only layer 2 relevant)
- [ ] Add top menu entries (projects, tools, apps, etc.) as search results  to hierarchy layer 1 and allow tabbing into them in order to list their submenus in layer 2 and 3
- [ ] Make checkboxes work for menus using them (e.g. `Extras >> Themes & Colors`)

FranzSw:
closes #187


FranzSw:
We already wrote some tests and added the submenu items to the first layer search.

What is still missing (and most of the complexity of this issue) is to add the submenu entries as items with child items in Algernon (where you can tab in)

bewee:
Still missing:
- ~~Add 4 window menu entries under line back in (probably best as separate actions on the window type?)~~ Done
- Maybe add actions toggle, enable and disable for checkbox menu entries? Edit: Ok, just change action name to toggle
- Fix tests
- ~~Find out why Colorful windows sometimes are not found???(?) (Edit: @FranzSw) -> The items are only found if the menu has been opened before. If you don't find the items, open the menu & submenu and the items are found again. why? ..~~ Done
- ~~Sometimes menu entries with submenu are not tabbable (same issue as above, every time the submenu items are not found)~~ Done

FranzSw:
~~bug: Unnamed can be found (the project name in the middle)~~ Done

FranzSw:
- ~~Discussion: Filtering the window items: Our idea was to filter / remove the items in the NamedWorldMenuEntry. Because they are dynamically generated, we would have to "inject" / pass a filter block. This feels a bit ugly, because it has some "zig-zag": The ALGWorldMenuType looks for the window-entry -> passes the filter block to the entry -> the entry filter the items. Any other ideas?~~ Yes, done.

Regarding the bug:
- We think it might have something to do that on every change in a menu a new menu instance is generated. We probably store the old instance somewhere in the code or whatever - these are then not up to date and don't have submenus anymore which is why the items are not found again until a menu item is opened.

bewee:
Bugs should now be fixed

packages/Algernon.package/ALGNamedWorldMenuEntry.class/instance/allEntries.st
```diff
@@ -1,12 +1,12 @@
...
+allEntries
 	| entries |
 	
 	entries := OrderedCollection new.
 
 	entries addAll: self menuEntries.
 	
 	entries addAll: (self menuEntries inject: OrderedCollection new into: 
-		[ :accumulator :entry | accumulator addAll: entry subMenus ]).
+		[ :accumulator :entry | accumulator addAll: entry allEntries; yourself ]).
```
niklasmohrin:
```suggestion
self menuEntries do: [ :entry | entries addAll: entry allEntries ].
```

packages/Algernon.package/ALGNamedWorldMenuEntry.class/instance/menuEntries.st
```diff
@@ -1,7 +1,9 @@
...
-		ifTrue: [^ OrderedCollection new ].
-		
-	^ self selectEnabled: self menuSubmorphs
\ No newline at end of file
+	self name = 'Windows' ifTrue: [^ OrderedCollection new ].
+
+	^ ((self menuSubmorphs
+		collect: [:morph | ALGNamedWorldMenuEntry newFrom: morph])
+		select: [ :entry | entry isValid])
+		select: [:entry | entry isEnabled]
```
niklasmohrin: Formatting is a bit odd, twice there is a space before the colon, once there isn't. Couldn't the two `select:`s be merged into one with `and:`, so that we don't create the intermediate collection?

```suggestion
^ self menuMorphs
    collect: [ :morph | ALGNamedWorldMenuEntry newFrom: morph]
    thenSelect: [ :entry | entry isValid and: [ entry isEnabled ] ]
```

packages/Algernon.package/ALGWorldMenuType.class/class/allEntries.st
```diff
@@ -0,0 +1,12 @@
...
+allEntries
+
+	|entries|
+	
+	entries := OrderedCollection new.
+	
+	entries addAll: self menuEntries.
+	entries addAll: (self menuEntries
+		inject: OrderedCollection new into: [:accumulator :entry | accumulator addAll: entry allEntries; yourself]).
+
```
niklasmohrin: See comment above

packages/Algernon.package/ALGWorldMenuType.class/class/menuEntries.st
```diff
@@ -1,15 +1,6 @@
...
-			entries addAll: (
-				(self extractMenuEntriesFrom: each)
-					collect: [ :entry | ALGNamedWorldMenuEntry newFrom: entry name: each name]) ].
-	
-	^ entries select: [:each | each morph isKindOf: MenuItemMorph]
-	
\ No newline at end of file
+	^ (self dockingBar submorphs
+		collect: [:morph | ALGNamedWorldMenuEntry newFrom: morph])
+		select: [:entry | entry isValid]
```
niklasmohrin: `collect:thenSelect:`

packages/Algernon.package/ALGWorldMenuType.class/class/menuEntriesNew.st
```diff
@@ -1,5 +1,14 @@
 algertem lists
 menuEntriesNew
 
-	^ (ALGNamedWorldMenuEntry newFrom: ActiveWorld mainDockingBars first)
-		subMenus
\ No newline at end of file
+	|menuEntries entries|
+	
+	entries := OrderedCollection new.
```
niklasmohrin: ~~unused~~ well, is is populated later, but that is never used

packages/Algernon.package/ALGWorldMenuType.class/class/menuEntriesNew.st
```diff
@@ -1,5 +1,14 @@
...
+	
+	menuEntries := (self dockingBar submorphs
+		collect: [:morph | ALGNamedWorldMenuEntry newFrom: morph])
+		select: [:entry | entry isValid].
+	entries addAll: menuEntries.
+
+	^ menuEntries
+		inject: OrderedCollection new into: [:accumulator :entry | accumulator addAll: entry subEntries; yourself]
```
niklasmohrin: Here, the `inject:into:` pattern makes sense, but you probably meant to use `entries` somehow

packages/Algernon.package/ALGNamedWorldMenuEntry.class/instance/allEntries.st
```diff
@@ -1,12 +1,12 @@
 accessing
-subMenus
+allEntries
 	| entries |
 	
 	entries := OrderedCollection new.
 
 	entries addAll: self menuEntries.
```
niklasmohrin: don't know if this is style, but it could also be

```suggestion
entries := OrderedCollection newFrom: self menuEntries.
```

niklasmohrin (commented):
don't know if you were looking for a review yet, but here you go 

edit: ah that was stupid, I only reviewed one commit :facepalm: some things still apply though

packages/Algernon.package/ALGNamedWorldMenuEntry.class/instance/enabledMenuSubmorphs.st
```diff
@@ -2,4 +2,4 @@ accessing
 enabledMenuSubmorphs
 	
 	"Some submorphs can't be disabled (e.g. line, ...)"
-	^ self menuSubmorphs select: [:menuItem | (menuItem class canUnderstand: #isEnabled) not or: [menuItem isEnabled]]
\ No newline at end of file
+	^ (self menuSubmorphs, self subMenuSubmorphs) select: [:menuItem | (menuItem class canUnderstand: #isEnabled) not or: [menuItem isEnabled]]
```
niklasmohrin, line 5: Do you mean `{ a . b }` to create an array? The `,` is string concat, right?

packages/Algernon.package/ALGWorldMenuType.class/class/menuEntries.st
```diff
@@ -1,16 +1,9 @@
...
-	
-	^ entries select: [:each | each morph isKindOf: MenuItemMorph]
-	
\ No newline at end of file
+	^ ((((self dockingBar submorphs
+		collect: [:morph | ALGNamedWorldMenuEntry newFrom: morph])
+		select: [:entry | entry isValid])
+		select: [:entry | entry hasEntries])
+		select: [:entry | entry name ~= 'Windows' ])
+		collect: [:entry | entry category: entry name]
```
niklasmohrin: last `collect` should be a `do`, right? Maybe something like
```smalltalk
^ (self make
          complicated
          stuff)
        do: [ ... ];
        yourself
```

better describes what is going on here? Or go with a temporary variable

packages/Algernon.package/ALGNamedWorldMenuEntry.class/instance/enabledMenuSubmorphs.st
TheGrayStone: I think this method is not even used anymore, is it? At least we removed the `subMenuSubmorphs` method...

niklasmohrin: That is ... interesting. CI would have (hopefully) caught that aswell

FranzSw: For ordered collections, the `,` merges them (instead of creating a 2D array), so I think it is correct.
If the method is not used anymore, we probably should remove it anyway :D

niklasmohrin: ohhh, yeah that actually makes sense, well then forget that comment

packages/Algernon.package/ALGNamedWorldMenuEntry.class/instance/allEntries.st
bewee: Not so sure whether that's really better. What do the others think?

bewee:
It's just another CI problem. All test pass on my machine.

bewee:
Concerning the latest commit: I think that it's a good idea not to cleanup the morphs after `updateMenu`, because
1. This destroys the `owner` field of submorphs. That's why we had to catch the exception dirtily in the first place.
2. When you open a dynamic world menu, then open algernon, the list gets cleared. This now no longer happens.

packages/Algernon.package/ALGActionVerbFilter.class/class/helpTexts.st
```diff
@@ -7,7 +7,6 @@ helpTexts
 		add: 'browse'->'Open a given class in the Browser.';
 		add: 'fileout'->'Generates a .st file in Resources of squeak image Folder';
 		add: 'open'->'Open an app, tool or morph';
-		add: 'saverepo'->'Saves to Repository with a working copy';
```
Olliwehr, line 10: I suppose this keyword is already outdated or is there another reason it was removed here?

packages/Algernon.package/ALGNamedWorldMenuEntry.class/instance/allEntries.st
FranzSw: I think it's way better :)

FranzSw:
`allEntries` exists on `ALGNamedWorldMenuEntry` and `ALGWorldMenuType` - are both necessary?

packages/Algernon.package/ALGNamedWorldMenuEntry.class/instance/allEntries.st
```diff
@@ -0,0 +1,10 @@
+accessing
+allEntries
```
FranzSw, line 2: `allEntries` exists on `ALGNamedWorldMenuEntry` and `ALGWorldMenuType` - are both necessary?

packages/Algernon.package/ALGNamedWorldMenuEntry.class/instance/allEntries.st
```diff
@@ -0,0 +1,10 @@
...
+	| entries |
+	
+	entries := OrderedCollection new.
+	
+	entries addAll: (self menuEntries inject: OrderedCollection new into: 
+		[ :accumulator :entry | accumulator addAll: {entry}, entry allEntries; yourself ]).
```
Olliwehr: Don't know if the code snippet here (or the one in ALGWorldMenuType) causes the following problem, but i think adding an extra `{entry}` causes the following situation:

When i search for a top-level World Menu Item (e.g. Extras or Apps) and tab into it (via the entries action), not only the entries of that top level are shown as sub-items, but i get an extra item for that `entries` action.

image

As far as i can tell clicking on that item doesn't really do sth. Quite strange, is that intended?

Olliwehr (commented):
It looks good overall, i got just two small things. I think while i reviewed @FranzSw already commented on one code snippet i commented on 😅

packages/Algernon.package/ALGNamedWorldMenuEntry.class/instance/allEntries.st
TheGrayStone: Yes, this is intended behavior to make it easier to search for entries whose name you know but are not sure in which menu they are.

packages/Algernon.package/ALGActionVerbFilter.class/class/helpTexts.st
FranzSw: Maybe we could leave 'save' as example for a menu item?

packages/Algernon.package/ALGNamedWorldMenuEntry.class/instance/allEntries.st
bewee: Ok, nice! Then let's just hope that the linter does not cry again because of the `do:`😅

packages/Algernon.package/ALGNamedWorldMenuEntry.class/instance/allEntries.st
bewee: Yes, one scans the docking bar entries (submorphs of docking bar) and the other one the submenu entries (submorphs of submenu of menu entry)
There is very little difference but quite a bit of duplication, so I'm not really satisfied with it either. But then again, SWT taught me to only deduplicate whenever I do the same thing for the **3rd** time 🙈

packages/Algernon.package/ALGNamedWorldMenuEntry.class/instance/allEntries.st
bewee: To my mind it's not intended, but fixing it would be too hard. It's not caused by this additional `{entry}` (that one is there for finding the top menu on the global search), since that's in `allEntries` but in the submenu view we only call `menuEntries`

bewee: We thought about opening the menu morph on this action in order for it to at least do something, but found it unnecessary

Olliwehr: @bewee ahh okay. Yeah, i also quickly talked with Nils about that issue. We resolved the discussion as we also thought that `fixing would be too hard` points it out. Maybe that is stuff for our last sprint before the submission, but for know i think its okay that way 👍

packages/Algernon.package/ALGNamedWorldMenuEntry.class/instance/subMenuIfAbsent..st
```diff
@@ -0,0 +1,6 @@
+accessing
+subMenuIfAbsent: aBlock
+	
+	^ (self morph class canUnderstand: #subMenu)
+		ifTrue: [ self morph subMenu ifNil: [^ aBlock value] ]
```
niklasmohrin:
```suggestion
		ifTrue: [ self morph subMenu ifNil: [aBlock value] ]
```

packages/Algernon.package/ALGWorldMenuType.class/class/allEntries.st
```diff
@@ -0,0 +1,10 @@
+algertem lists
+allEntries
+
+	| entries |
+
+	entries := self menuEntries inject: OrderedCollection new into:
+		[ :accumulator :entry | accumulator addAll: {entry}, entry allEntries; yourself].
```
niklasmohrin, line 7: hm, I am still in favor of the version from https://github.com/hpi-swa-teaching/Algernon-Launcher/pull/221/commits/94e4cdd1ae54f685da9af3a7ef4ceddb817909f2 until there is a `Collection>>flatten`. This whole `inject:into:` business with the concatenation of the collections feels so noisy in comparison to `self menuEntries do: [...]` and I don't really see why we should kneel to some odd linter rule that doesn't want us to write more concise code... :man_shrugging:

If you really insist on this version, I am okay with `inject:into:`

packages/Algernon.package/ALGActionVerbFilter.class/class/helpTexts.st
niklasmohrin: I think we removed it because it was already outdated and just happened to see it still here

packages/Algernon.package/ALGNamedWorldMenuEntry.class/instance/allEntries.st
```diff
@@ -0,0 +1,4 @@
+accessing
+allEntries
+	^ self menuEntries inject: OrderedCollection new into: 
+		[ :accumulator :entry | accumulator addAll: {entry}, entry allEntries; yourself ]
```
niklasmohrin, line 6: Here, I would prefer `inject:into:` too :+1: I have another suggestion to get rid of the array concat though, what do you think of this change?
```suggestion
	^ self menuEntries
		inject: OrderedCollection new
		into: [ :accumulator :entry |
			accumulator
				add: entry;
				addAll: entry allEntries;
				yourself ]
```

packages/Algernon.package/ALGNamedWorldMenuEntry.class/instance/isValid.st
```diff
@@ -0,0 +1,4 @@
+accessing
+isValid
+
+	^ (self morph isKindOf: DockingBarItemMorph) or: [self morph isKindOf: MenuItemMorph]
```
niklasmohrin: I feel like the word "valid" is a bit broad here, either we should add a comment how an instance could be invalid, or come up with a more precise term. Or I have overread something. That is also a possibility

packages/Algernon.package/ALGNamedWorldMenuEntry.class/instance/selector.st
```diff
@@ -0,0 +1,4 @@
+accessing
+selector
+	^ (self morph selector
+		ifNil: [ String new ]) asString
```
niklasmohrin:
```suggestion
selector
	^ (self morph selector
		ifNil: [ String empty ]) asString
```

packages/Algernon.package/ALGWorldMenuType.class/instance/determineActionName.st
```diff
@@ -0,0 +1,13 @@
+actions
+determineActionName
+	
+	self typeObject hasEntries ifTrue: [ ^ 'entries' ].
+	self class actionNameSelectors
+		detect: [ :selectorString | self typeObject selector startsWith: selectorString]
+		ifFound: [ :element | ^ element ].
```
niklasmohrin:
```suggestion
		ifFound: [ :selectorString | ^ selectorString ].
```

packages/AlgernonTests.package/ALGTestTypes.class/instance/testSubmenuCheckBoxes.st
```diff
@@ -0,0 +1,15 @@
...
+	colorfulWindowsName := 'Colorful Windows'.
+	colorfulWindowsInitialValue :=  Model useColorfulWindows.
+	self runQuery: colorfulWindowsName.
+	self assert: (self resultsIncludeName: colorfulWindowsName).
+	self doActionForItem: (self items first).
+	self deny: colorfulWindowsInitialValue equals: Model useColorfulWindows.
+	
+	Model useColorfulWindows: colorfulWindowsInitialValue
+	
+
```
niklasmohrin, line 15: if we happen to touch this method again, we should remove the trailing empty lines, it's not a must for now though (this is in other test methods too)

packages/Algernon.package/ALGWorldMenuItem.class/instance/icon.st
```diff
@@ -2,5 +2,5 @@ icon
 icon
 
 	^ self class mapping
```
niklasmohrin: I think we should come up with a better name than "mapping", at least something like `iconMapping` or so

packages/Algernon.package/ALGNamedWorldMenuEntry.class/instance/selector.st
```diff
@@ -0,0 +1,4 @@
+accessing
+selector
+	^ (self morph selector
```
niklasmohrin, line 3: not every morph understands `#selector`, **imo** we should either rename the instance variable to a more specific type, add a comment here or check whether it understands the message

niklasmohrin (changes requested):
I think its a bit odd that the window actions are separated from the rest now, but if we would have to do the check in the general code anyways, I am okay with that

packages/Algernon.package/ALGNamedWorldMenuEntry.class/instance/allEntries.st
niklasmohrin: this was really hard to grasp when laura and I started working last week, maybe we could add a short comment that outlines the situation or just deduplicate? :smile:

I think that a general helper method for traversing / collecting the submorphs on the one that is not an `ALGType` would be great, you can decide though

packages/Algernon.package/ALGNamedWorldMenuEntry.class/instance/subMenuIfAbsent..st
FranzSw: Isnt it then just `aBlock` (`aBlock` == `[aBlock value]`)?

niklasmohrin: yes, but I think there is yet another lint that forbids using anything but a new block literal after `ifTrue:` etc.

Personally, I, too, would prefer just writing `aBlock`, although I can see how this lint makes sense (unlike the one about `do:`). Should we disable the lint?

packages/Algernon.package/ALGNamedWorldMenuEntry.class/instance/selector.st
TheGrayStone: We **know** that morph is a MenuItemMorph, which is easily visible from the code, so I think it's okay here.

packages/Algernon.package/ALGNamedWorldMenuEntry.class/instance/allEntries.st
TheGrayStone: We have decided to keep the concat because we think it is very elegant and reads well.

niklasmohrin (approved):

PR #222, opened 2021-07-07
Spike for 'context sensitive actions'

### LINKED ISSUE #208: Investigate context-sensitive actions
As an Algernon enthusiast, I want to find context-sensitive actions, in order to speed-up my workflow.

## Acceptance Criteria

- [ ] Research possible solutions & issues
- [ ] Build Spike if feasible

TheGrayStone:
closes #208 
-------------------------------------
This is based on the reworked menu item structure


niklasmohrin (commented):
Neat, this works with only this small change :+1:

There are some things that I noticed though (mostly no must-haves for this spike):
- the keyboard focus is reset to the end of the input field after a second or so, so moving to the beginning of the input is virtually impossible. We've got to come up with something better than just `moveCursorToEnd` (same goes for say selecting some text)
- the above problem also means that the text is unselected very quickly when you open algernon, so you usually start appending to your last query instead of overwriting it
- when activating algernon while hovering over an entry of a right-click-menu, the menu disappears, but algernon doesn't get keyboard focus, so neither querying nor closing with esc work. one can click into the input field to fix that glitch
- I think we want to have separate icons for these actions. Otherwise, lots of results have the world icon. in my opinion, having these results as their own type seems appropriate in any case
- do we know that all of the context menus consist of `MenuMorph`s?
- results from the second page (after you click "more") are not included, for example "inspect subinstances". with the current approach, I don't really see us finding these any time soon though (and it isn't really a must, not even for the first release of this feature imo)
- talking about the approach: while it is very direct and effective, I wonder if we could only "fetch" the results from the normal provider that is used by the browser (<= really assuming some architecture of the system browser right here). "clicking around" in the world on every keypress doesn't sound very safe. Are there any scenarios where right-clicking actually really _does_ something meaningful (more than the example above)? That would be pretty bad

TheGrayStone:
> Neat, this works with only this small change 👍
> 
> There are some things that I noticed though (mostly no must-haves for this spike):
> 
> * the keyboard focus is reset to the end of the input field after a second or so, so moving to the beginning of the input is virtually impossible. We've got to come up with something better than just `moveCursorToEnd` (same goes for say selecting some text)
> * the above problem also means that the text is unselected very quickly when you open algernon, so you usually start appending to your last query instead of overwriting it
> * when activating algernon while hovering over an entry of a right-click-menu, the menu disappears, but algernon doesn't get keyboard focus, so neither querying nor closing with esc work. one can click into the input field to fix that glitch
> * I think we want to have separate icons for these actions. Otherwise, lots of results have the world icon. in my opinion, having these results as their own type seems appropriate in any case
> * do we know that all of the context menus consist of `MenuMorph`s?
> * results from the second page (after you click "more") are not included, for example "inspect subinstances". with the current approach, I don't really see us finding these any time soon though (and it isn't really a must, not even for the first release of this feature imo)
> * talking about the approach: while it is very direct and effective, I wonder if we could only "fetch" the results from the normal provider that is used by the browser (<= really assuming some architecture of the system browser right here). "clicking around" in the world on every keypress doesn't sound very safe. Are there any scenarios where right-clicking actually really _does_ something meaningful (more than the example above)? That would be pretty bad

It's a spike...

PR #223, opened 2021-07-29
Fix 'Evaluate text field' when no keyword is given (= / do)

### LINKED ISSUE #172: Bug: Evaluate Text Field ignores first two characters
## Reproduction
1. Open Algernon
2. Type in `Object new` (or `2+2`)
3. Select Evaluate text field

(if you prepend two whitespaces, results are correct)

## Issue
First two characters are ignored, which results in wrong results or errors 

e.g. Object new -> Error: `ject is Undeclared`
1+2 -> 2

Olliwehr:
closes #172

While those two prefixes really indicate that the following should be interpreted as an expression, the 'evaluate text field' option is also available when none of those two is given. In this case, the two first characters of the actual command got cut, resulting in weird behaviour.


bewee (approved):
Way better! 👍

PR #224, opened 2021-07-31
Rename worldmenu to dockingbar

bewee:


bewee:
Note: There are still some artifacts of static / dynamic which we should probably eliminate as well. Do we want to do this within this PR or in a separate one?

FranzSw:
@bewee ~~Let's do it here - it is related and just refactoring~~ Changed my mind, the changelog is already big :D

If you have time we could also do it together some time

FranzSw (approved):

PR #228, opened 2021-08-04
Fix coveralls link in README.md

niklasmohrin:
We can also put similar patches in this PR and make the title more general


FranzSw (approved):
Finally! :)

PR #226, opened 2021-08-03
Small fixes from our presentation

TheGrayStone:
Small syntactic fixes before handing in the project...

Related to https://github.com/hpi-swa-teaching/Algernon-Launcher/projects/1#card-63890788

PR #229, opened 2021-08-04
Coverage

niklasmohrin:
Feel free to push all the fixes about coverage to this branch :)


niklasmohrin:
FYI: During development, windows have been closed unexpectedly. Programming is hard

packages/AlgernonTests.package/ALGMockedObject.class/properties.json
```diff
@@ -0,0 +1,15 @@
...
+	"category" : "AlgernonTests",
+	"classinstvars" : [
+		 ],
+	"classvars" : [
+		 ],
+	"commentStamp" : "",
+	"instvars" : [
+		"mockedMethods",
+		"calls" ],
+	"name" : "ALGMockedObject",
```
Olliwehr, line 11: Would it be better to also rename your new mocked objects to use "Fake" instead of "Mocked" (as the other mocked things in the testing package follow this naming convention)? But that is just nitpicking tbh :D

packages/AlgernonTests.package/ALGTestAction.class/instance/testWindowOperations.st
```diff
@@ -0,0 +1,14 @@
+testing
+testWindowOperations
```
Olliwehr: I suggest putting those two new test cases in the `ALGTestTypes` test class (category `testing window type` maybe?) as at least for me that is what we do here (rather than the the other tests in `ALGTestAction` that mostly test the creation of Algeractions). What's your opinion on that?

packages/AlgernonTests.package/ALGTestAction.class/instance/testWindowOperations.st
```diff
@@ -0,0 +1,14 @@
...
+
+	| instance |
+	instance := ALGMockedWindowOperationsType new.
+	instance
+		closeAllButWorkspaces;
+		closeAllWindows;
+		closeWithoutChanges;
+		collapseAllWindows.
+	self assert: 3 equals: instance mockedDockingBar calls keys size.
```
Olliwehr: The current coverage report states that your newly introduced indirections for the docking bar/the world in the WindowType and the WindowOperationsType aren't covered. 

I question if that is still the case if we use `dockingBar` instead of `mockedDockingBar` here. That won't make a difference in behaviour as you overwrote the dockingBar shorthand in the MockedWindowOperationsType with this indirection, right? 😄 

If this will lead to a coverage of the `dockingBar` shorthand then the same could be applied to the `world` shorthand here and the `dockingBar` one in the WindowType. Alltogether i think  this would bring us back over 99% again 🎉

Olliwehr:
Your changes are looking good overall!

But i got no idea why merging the current development into this branch failed all tests. All checks now time out after 15mins?

packages/AlgernonTests.package/ALGTestAction.class/instance/testWindowOperations.st
niklasmohrin: Well, I didn't think of that :laughing: 

I wouldn't change `mockedDockingBar` to just `dockingBar` here, because it is not what wanted to express. The test doesn't _really_ care about docking bars or in which way the window operations get them, it just wants to check up on the mock object. I think if we changed this to `dockingBar`, we would also run into the method of the mocked object and not into the one that is currently not covered.

I think we could add a test that verifies the interface of whatever the `dockingBar` and `world` methods return. Maybe get the object and then assert that it can understand the messages that we need it to understand (`closeAllWindows` etc.). This way, we would have the coverage and the test would be meaningful (maybe we change the docking bar at some point? don't know if that is possible :D)

What do you think?

packages/AlgernonTests.package/ALGMockedObject.class/properties.json
niklasmohrin: Hm, I think when coming up with the names, I felt that classes like `FakeKeyboardEvent` are more about gaining access to internals of existing classes whereas I wanted the `ALGMockedObject` to really be just a dumb call-recorder. I guess we can rename them all to "Fake" for consistency, but then we should think of another name than "FakedObject" or add a class comment

niklasmohrin:
Maybe github or smalltalkci is having a problem? Tests aren't even starting and we didn't touch anything special, did we? :man_shrugging:

packages/AlgernonTests.package/ALGMockedObject.class/properties.json
Olliwehr: Ahh okay, yeah good point. Then i think its better to go with "Mocked" here as it is now 👍

packages/AlgernonTests.package/ALGTestAction.class/instance/testWindowOperations.st
Olliwehr: I see your point. Didn't thought about whether that would be really meaningful. Was more like an approach based on pure coverage 😄 

Anyways, the 2 latest commits got us back over 99% regardless :D

niklasmohrin:
hm, neither smalltalkci nor setup-smalltalkci had a release, this is odd

packages/AlgernonTests.package/ALGMockedObject.class/properties.json
niklasmohrin: I renamed the two algtypes but kept the MockedObject for now

niklasmohrin:
okay, we are at 99.57% now, I guess this should be enough (badge says 100% :tada: )

Question remains: Why are baseline tests stuck? The coverage reporter runs all normal tests in 3 mins, but everything else times out on the one baselineofalgernontests test :thinking:

Olliwehr:
Hmm, could it be related to the introduction of `self ensureInternetConnection` in #226 in the BaselineTest? But don't know why accesing google.com should cause a timeout, but if, then i think that is problem with github actions and the connection of their VM's... 🤔

Edit: its just a guess tho, just seen that those changes we already merged when you created that branch here

TheGrayStone:
On my local machine the baseline tests also take forever to complete, even crashing the vm sometimes. 🤔

niklasmohrin:
Let's consult the slack?

niklasmohrin:
Another idea: lets rerun CI on the commit where it last worked to check if it is a toolchain or a code problem (I cant right now, just on mobile)

TheGrayStone:
image
Maybe this explains it?

Although https://www.githubstatus.com/ says everything is ok. 🤔

niklasmohrin:
Update:

> Investigating - We are investigating reports of degraded performance for GitHub Actions. 
Aug 5, 14:09 UTC

packages/AlgernonTests.package/ALGTestTypes.class/instance/testWindowOperations.st
```diff
@@ -0,0 +1,14 @@
...
+	instance := ALGFakeWindowOperationsType new.
+	instance
+		closeAllButWorkspaces;
+		closeAllWindows;
+		closeWithoutChanges;
+		collapseAllWindows.
+	self assert: 3 equals: instance mockedDockingBar calls keys size.
+	self assert: 1 equals: instance mockedWorld calls keys size.
+	instance mockedDockingBar calls values do: [ :coll | self assert: 1 equals: coll size ].
+	instance mockedWorld calls values do: [ :coll | self assert: 1 equals: coll size ]
```
bewee: Maybe consider giving `coll` a more descriptive name?

packages/AlgernonTests.package/ALGMockedObject.class/instance/recordCall..st
```diff
@@ -0,0 +1,7 @@
+mocking
+recordCall: aMessage
+
+	(self calls
+		at: aMessage selector
+		ifAbsentPut: [ OrderedCollection new ])
+			add: aMessage
```
bewee: I find this a little odd to read, maybe we could add a temporary?

packages/AlgernonTests.package/ALGTestTypes.class/instance/testWindowTypeDockingBarMock.st
```diff
@@ -0,0 +1,10 @@
+testing window type
+testWindowTypeDockingBarMock
+
+	| instance |
+	instance := ALGFakeWindowType new.
+	instance
+		closeAllWindowsLikeThis;
+		closeAllWindowsButThis.
+	self assert: 2 equals: instance mockedDockingBar calls keys size.
+	instance mockedDockingBar calls values do: [ :coll | self assert: 1 equals: coll size ]
```
bewee: Same here with `coll`

bewee (approved):
Great stuff!

bewee:
Ah, okay, so now tests get executed here as well. Weird.

@niklasmohrin @Olliwehr There really seems to be a serious problem with the code from this branch that leads to infinite test execution time. Maybe something's wrong with one of those new tests? 🤔

TheGrayStone (approved):
Yes, finally!

niklasmohrin (commented):
I really like what you have done here

PR #231, opened 2021-08-05
Cleanup and refactor tests

bewee:
- Add `assert: aBlock opensNewWindow: name` and similar helpers and use them wherever possible. Open question: Some forms of the helpers added may not be used, should we then remove them?
- Add similar helpers for `SystemDialog`s and use them
- Replace `className` variable in `ALGTestClassReload` which never changed with a constant method
- Remove some unnecessary (now already handled by the ALGTest base class) manual setup and teardown functionalities implemented by some tests
- Many small thingies I stumbled upon, like trailing dots


bewee:
One more thing: Removed the test `testDynamicItemsOnlyOnDefaultSearchLevel` because neither me nor @TheGrayStone recognize its purpose

Olliwehr:
@bewee regarding the `testDynamicItemsOnlyOnDefaultSearchLevel`:

Take a look at `ALGItemProvider >> searchAllItemsFor:` There we only enrich the found static items (so normal class or method results for example) with the dynamic items (so for example things from the docking bar) if we are on the default search level. Thats quite an important thing to do, as when you increase the search level (by navigating into a class result for example) you _only_ want to see results regarding the parent item you just tabbed into (and not some dynamic items!).

And exactly that behaviour was covered from the test you now removed... 😄

bewee:
Ah, right, this does make sense, but wasn't really that obvious because of the (simply wrong) test name and misleading comments in the code 🙈 
Added the test back in, now with a different name which hopefully indicates its purpose more clearly :D

Edit: @Olliwehr Thinking about it slightly longer, I feel like maybe it should be two separate tests? One `testDynamicItemsNotInForeignChildLists` and one `testDynamicItemsOnDefaultSearchLevelFound`. What do you think?

Olliwehr:
@bewee nah, as a foreign child list is definitively not on the default search level, it would be quite the same test or what do you imagine the second one to be like? :D

Olliwehr (approved):
I think future teams will absolutely _love_ the newly introduced helper methods regarding the window opening stuff. 😄 Nice work!

bewee:
> @bewee nah, as a foreign child list is definitively not on the default search level, it would be quite the same test or what do you imagine the second one to be like? :D

I thought about something like

```
testDynamicItemsOnDefaultSearchLevelFound

	self runQuery: 'Transcript clear'.
	self assert: ((self itemMorphs 
					reject: [:each | each item isNil])
					anySatisfy: [:each | each item algerType isKindOf: ALGDockingBarType ])
```

```
testDynamicItemsNotInForeignChildLists

	self runQuery: 'Monticello' withCursorAt: 'Package'.
	self algInstance showChildList.
	self deny: ((self itemMorphs 
					reject: [:each | each item isNil])
					anySatisfy: [:each | each item algerType isKindOf: ALGDockingBarType ]) 
```

They just feel like different things to me.
Although I would be fine with leaving it as it is as well 😅

PR #248, opened 2022-05-16
Fixed test cases for squeak 6.0 trunk version

yannik-dittmar:
Found out what went wrong for the new squeak trunk version.

- Deprecated message `#startsWith:`
- Changes to the squeak submenu


packages/AlgernonTests.package/ALGTestTypes.class/instance/testOpenSubmenuEntries.st
```diff
@@ -1,12 +1,16 @@
...
 	
 	fontSizeName := 'Increase Font Size'.
+	"Item name changed in squeak v6.0"
+	newFontSizeName := 'increase font size (+)'.
 	self runQuery: fontSizeName.
-	self assert: (self resultsIncludeName: fontSizeName).
\ No newline at end of file
+	self assert: (
+		(self resultsIncludeName: fontSizeName) or: 
+		[self resultsIncludeName: newFontSizeName]).
```
gwauge, line 16: @Kryptolyser, ist da das "or" wichtig oder kann man das alte (`fontSizeName`) auch einfach weglassen?

gwauge (approved):
Du alter Hacker....
LGTM!

packages/AlgernonTests.package/ALGTestTypes.class/instance/testOpenSubmenuEntries.st
yannik-dittmar: No, the `or` is important, because we have different submenu entries depending on the squeak version. If you would remove the old `fontSizeName` it would fail on Squeak 5.3 CI.

PR #250, opened 2022-05-19
Refactor: Execute linter rules all at once to save time

yannik-dittmar:
Save time on linting by executing all linter rules with all code methods at the same time. 
This prevents the SLTestRunner from having to load the necessary plugin every time.


matveysiva (approved):
LGTM
MfG Frederic

PR #253, opened 2022-05-30
Fix: Remove algernon instances after tests are finished.

### LINKED ISSUE #245: Algernon instances used in tests aren't properly disposed
# User story
As an Algernon dev, I would like to run tests without breaking my working version of Algernon, in order to speed up development time.

# Issue
when one runs Algernons test suite (or selected __UI__ tests), these tests apparently open up new Algernon instances to perform their assertions. However it seems they do not properly dispose of the Algernon instance. This results in mutiple Algernon windows being overlayed on top of each other which can only be closed by pressing the close key (i.e. `esc`) dozens of times or holding it for an extende period of time.
This isn't an issue for CI runs, since the CI image gets deleted. It is however very annoying in the development process.

# Reproduction
- run some or all test cases, that include UI assertions
- open Algernon as usual
- observer the background being opaque white, instead of partially transparent
- try closing Algernon

# Proposed fix
- global teardown (if possible) that closes all the instance
- shouldn't close the actual instance on the system

# Temporary workaround
- run `ALGCore allInstances do: [:item | item delete]` in your workspace to close __all__ instances
- open up `Preferences >> Algernon` and re-enable it
- back to normal!

gwauge:
- Resolves #245

Co-authored-by: Georg Auge <georg.auge01@gmail.com>

### DIFF
--- a/packages/AlgernonTests.package/ALGTest.class/instance/tearDown.st
+++ b/packages/AlgernonTests.package/ALGTest.class/instance/tearDown.st
@@ -1,8 +1,8 @@
 running
 tearDown
 	self algInstance deactivate.
-	self algInstance delete.
 	(self currentSystemWindows difference: self previouslyOpenedWindows)
 		do: [:each | each delete].
+	ALGCore allInstances do: [:item | item delete].
 	self restorePreferences.
 	Algernon currentInstance: self previousAlgernonSingleton.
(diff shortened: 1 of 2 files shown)


yannik-dittmar (approved):
Looks hacky to me... But if it works it should be good enough for the test

yannik-dittmar (approved):
I don't know if this is intentional, but every time you run a test, Algernon is enabled, even if you had disabled it before starting the test. However, since I can't think of any situation where something like this would cause problems, I approve this PR.

PR #260, opened 2022-06-09
Equation solving

### LINKED ISSUE #244: Equation solving
As a user, I want to directly see the result of an equation I entered in order to minimize my interactions with Algernon.

matveysiva:
- Resolves #244


packages/Algernon.package/ALGExpressionType.class/instance/getDisplayedResult.st
```diff
@@ -0,0 +1,11 @@
+accessing
+getDisplayedResult
+
+	|command equationsResult|
+	
+	command := self getCommand asString.
+	
+	self isEquation
+		ifTrue: [equationsResult := Compiler evaluate: command for: self logged: false.
```
yannik-dittmar: Maybe a try catch here to prevent unwanted error messages from showing up and making Algernon unusable.

yannik-dittmar (changes requested):
Syntax error when using `^`:
image

Why is the result down at the bottom when putting two spaces between the numbers? Maybe we could force the expression item to be at the top when an equation is detected?
image

Divide by zero error (solution idea stated in file comment):
image

LinqLover:
> Syntax error when using `^`:
> image

FYIO, mathematical power in Smalltalk is expressed via `**`:

```smalltalk
2**3 "8"
```

yannik-dittmar:
> > Syntax error when using `^`:
> > image
> 
> FYIO, mathematical power in Smalltalk is expressed via `**`:
> 
> ```smalltalk
> 2**3 "8"
> ```

Thank you for your info, but then the regex shouldn't match this as an equation.

packages/Algernon.package/ALGExpressionType.class/class/equationPattern.st
```diff
@@ -1,4 +1,4 @@
 icon
 equationPattern
 
-	^ '( *[+\-]? *\d+)( *[+\-*\/^]? *\d+)* *'
\ No newline at end of file
+	^ '( *[+\-]? *\d+)( *[+\-*\/]? *\d+)* *'
```
yannik-dittmar: Potentiating doesn't throw errors, but still doesn't work using `**`.

packages/AlgernonTests.package/ALGTestTypes.class/instance/testAssertEquationPattern.st
```diff
@@ -3,5 +3,4 @@ testAssertEquationPattern
 
 	self assert: ['1+ 1 + 1' matchesRegex: ALGExpressionType equationPattern.].
 	self assert: ['-1 - 11' matchesRegex: ALGExpressionType equationPattern.].
-	self assert: ['2^ 3' matchesRegex: ALGExpressionType equationPattern.].
 	self assert: ['2/2' matchesRegex: ALGExpressionType equationPattern.].
```
yannik-dittmar, line 6: Maybe add another test for divide by zero?

yannik-dittmar (changes requested):
Syntax error when placing a space between two numbers:
image

packages/Algernon.package/ALGExpressionType.class/class/equationPattern.st
matveysiva: We removed it intentionally because it would require a much more complex RegEx. For now we reduced equation solving to the basic operators only

yannik-dittmar (approved):
LGTM

PR #261, opened 2022-06-09
Update determineActionName.st

EnnioEnnio:
Update determineActionName.st to work with latest Trunk Version.


gwauge:
Thank you for your contribution! We'll review it in due time. <3

